PR #5, opened 2019-10-20
Birds

mirisuzanne:


content/birds/miriam.md
```diff
@@ -42,3 +43,37 @@ writing, music, and visual art.
 [jss]: https://www.sitepoint.com/premium/books/jump-start-sass
 [speaking]: /services/speaking/
 [dev]: https://blog.cssdevconf.com/slides-and-resources-from-css-dev-conf-2017-new-orleans-8e2a5edb06b0#2b07
+
+------
+
+## Social
```
jgerigmeyer: How do you decide when to use headings here (and not on my page)?

mirisuzanne: it wasn't a decision - I wasn't thinking of this as final design/text but proof of tech. Can we list the things, and does it work right in two use-cases?

mirisuzanne: (trying to get quickly to a point where we've proved all the architectural patterns, and more people can do the work of cleaning up content/designs.

PR #7, opened 2019-11-27
Add pages.withData to filter a collection for page data

mirisuzanne:
Hedgehog with a hat

## Description
Attempt to abstract a common collection filter


src/filters/pages.js
```diff
@@ -13,26 +13,30 @@ const fromCollection = (collection, page) => {
...
-
 const titleSort = (collection) =>
   collection.sort((a, b) => a.data.title - b.data.title);
 
+const withData = (collection, key, value) =>
+  collection.filter((page) => {
+    const data = page.data[key];
+
+    if (data) {
+      if (data === value || value === null) {
```
duggiemitchell: If you are checking for null _and_ undefined, I think this line can be less terse like so: ```(data === value || !value) ``` otherwise if just null then you could use the regular equality operator `(data === value || value == null)`

PR #16, opened 2020-01-14
Macro & Filter Documentation

mirisuzanne:
Hallucigenia?

## Description
I used [Doxray](https://github.com/himedlooff/doxray) to parse inline documentation in both njk and js files, and create 11ty data from the results - then display them in the `/sample/` section of our site.

This isn't about how the results *look* (we still nee to build out the page template and typography styles) - but I'd love review on the general approach, and also on the JS. I bet there might be some better ways to handle some things?

[Site Preview »](https://deploy-preview-16--oddleventy.netlify.com/sample/)


content/_data/filters.js
```diff
@@ -0,0 +1,60 @@
...
+        data.title = data.info ? data.info.label : data.name;
+        data.patterns = docs.patterns.filter(
+          (pattern) => pattern.category !== 'file',
+        );
+
+        js.push(data);
+      });
+  });
+
+  return js;
```
wlonk: This can't possibly be right; the empty array `js` will be returned before the callback in `readdir` finishes, surely? Is this working for you, and if so, how?

content/_data/macros.js
```diff
@@ -0,0 +1,60 @@
...
+        data.title = data.info ? data.info.label : data.name;
+        data.patterns = docs.patterns.filter(
+          (pattern) => pattern.category !== 'file',
+        );
+
+        njk.push(data);
+      });
+  });
+
+  return njk;
```
wlonk: Ditto the above comment.

content/_data/filters.js
mirisuzanne: hmmm, it is working for me (and seemingly on the server), so I guess I need advice on what the fix would be? I really don't understand sync/async stuff yet - which I imagine is the category of issue here?

wlonk: So, the callback in `readdir` happens at an unspecified future point; I guess it's mutating the `js` array, which is being passed by reference, so as long as any code _using_ that array operates on it after that unspecified future point, it's OK, though kinda by accident. I think the Right Thing™ is to return a promise and resolve that promise when the callback finishes, and have anything operating on this array happen in a `then` (or `await`) off that promise.

mirisuzanne: @wlonk are you willing to make that change? I'm not clean onn how promises work.

wlonk: Yes, I absolutely am, but I want to make sure that the calling context can handle that, too.

content/_layouts/docs.njk
```diff
@@ -0,0 +1,92 @@
...
+    <h4>Todo:</h4>
+    {{ pattern.todo | md | safe }}
+  {% endif %}
+{% endmacro %}
+
+{% macro show(
+  patterns
+) %}
+  {% for pattern in patterns %}
+    <article id="{{ pattern.slug }}">
```
stacyk: Should we bump all these h3s and h4s up since there are no H2s on these doc pages?

stacyk: OH, I take that back. some pages don't have an H2 until "see also" because the content on the page doesn't have a "category" assigned like the `.../sample/macros/contact/` page, but `...sample/macros/embed/` has categories that become H2s like "Patterns", "Embed" "Layout". Should we always add a category even if it is a shared one by all items on that page? Or should we look for a category and if there isn't one, make the heading level one up from the pages with categories? That option seems more complicated.

content/sample/type.md
```diff
@@ -59,20 +66,18 @@ Consequat scelerisque egestas ut mollis feugiat sit velit per cubilia, gravida d
...
+
+{{ quotes.from(collections.all, '/work/metadeploy/', 'extension') }}
 
 Or in a grid:
 
-{{ quotes.grid(collections.all | pageData(
-  '/work/timedesigner/',
-  'press'
-)) }}
+{{ quotes.from(collections.all, '/work/timedesigner/') }}
```
stacyk: These examples are helpful. Thank you.

src/filters/time.js
```diff
@@ -68,16 +79,41 @@ const formatDate = (date, format) => {
...
+    note: See js file for a list of available date formats
+*/
 const getDate = (date, format) => {
   date = typeof date === 'string' ? new Date(date) : date || now;
   return format ? formatDate(date, format) : date;
 };
 
+/* @docs
+label: rssDate
+category: rss
```
stacyk: Should this read as `RSS` on the page? It must be going through a capitalize filter instead since it outputs to `Rss`

src/filters/utils.js
```diff
@@ -1,50 +1,160 @@
...
+const items = (list, start, end) => list.slice(start, end);
+
+/* @docs
+label: styles
+category: formatting
+note: |
+  Convert an object of property-value pairs
+  into valid CSS for an inline-style.
+example: |
+  <div style="{{ style_dict | styles }}">
```
stacyk: Do we use this anywhere? I imagine it would be helpful for setting certain Custom Properties

content/_includes/quotes.macros.njk
```diff
@@ -63,23 +83,95 @@
...
+  {%- else -%}
+    {{- grid(data) if data else none -}}
+  {%- endif -%}
 {%- endmacro %}
 
-{% macro _cite(
+
+{# @docs
+label: grid
+category: from data
```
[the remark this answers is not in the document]
mirisuzanne: yeah, these are (mostly) not required, and just used to group macros inside a longer file. The only special case is the `file` category used to describe the page.

content/_includes/toggle.macros.njk
```diff
@@ -1,10 +1,48 @@
...
+    type: string
+    note: Must match the `id` passed to `toggle.target()`
+  expanded:
+    type: boolean
+    default: 'false'
+    note: Current toggle status
+  button:
+    type: boolean
+    default: 'true'
+    note: Add `button` ARIA role, if the underlying element is not a button
```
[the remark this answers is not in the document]
mirisuzanne: Not yet. We also don't use these macros yet (but I think we will). 😛 I copied these over, and then changed the default from `false` to `true` for that reason. I'd love to get rid of that option, and just let it be handled in `attrs`  if we do need it?

content/_layouts/docs.njk
mirisuzanne: I'm open to either of these options. There's already special handling for no categories, so that one isn't actually more complicated. Don't have an idea for what  it should say, though. Maybe just "Macros"?

src/filters/utils.js
mirisuzanne: Yeah, that was the idea. I've mostly been using it in the `embed` macros, as a way to pass in variations.

content/_layouts/docs.njk
stacyk: We lost the header I think now: 
missing-oddbird-header-local-vs-preview

src/filters/utils.js
mirisuzanne: yeah, we use it a few places for exactly that :)

mirisuzanne:
[I want documentation of Filters & Macros](https://trello.com/c/5pUzylFs/30-i-want-documentation-of-filters-macros)

content/_data/filters.js
```diff
@@ -0,0 +1,58 @@
...
+            slug: file.slice(0, file.indexOf('.')),
+            title: info ? info.label : name,
+            patterns: docs.patterns.filter(
+              (pattern) => pattern.category !== 'file',
+            ),
+          };
+        });
+      resolve(js);
+    });
+  });
```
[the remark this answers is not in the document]
mirisuzanne: I think we have generally been using the latest, so we might as well? Is that an update you could  make @duggiemitchell ?

content/_includes/page/meta.njk
```diff
@@ -1,16 +1,19 @@
...
 } if elsewhere else none %}
 
+{% set sub = renderData.sub or sub %}
+
 {% if sub or public_tags or elsewhere or author %}
   <div id="meta">
     {% if sub or dates %}
       <p class="subtitle">
-        {{ sub | mdInline('amp') | safe }}
+        {{ sub | mdInline | safe }}
```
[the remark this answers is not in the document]
mirisuzanne: `amp` is one of several typography helpers we have, to ensure we use our fancy ampersand. I had a special-case for applying it without the other typography settings, but decided to update `mdInline` so it would handle this automatically.

src/filters/events.js
```diff
@@ -69,14 +115,21 @@ const fromYAML = (page) => {
...
 const fromCollection = (collection) => {
   const events = [];
 
   collection
     .filter((page) => pages.isPublic(page))
     .forEach((page) => {
       if (hasEvents(page)) {
-        Array.prototype.push.apply(events, fromYAML(page));
+        Array.prototype.push.apply(events, fromData(page));
       }
```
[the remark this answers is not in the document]
mirisuzanne: These are copied over from my site and will likely change a lot. On my site, I have two types of "events": some are individual pages (like an article or one-off theater piece), and some are in the YAML data of a page (like one talk being given multiple times at different venues). This cycles through all the pages in a collection, and returns an array of events from both sources.

I think we'll want something similar, but a bit more refined,for our resources page. I haven't gotten to those details yet… I'll need more review once I get there. :)

src/filters/events.js
```diff
@@ -107,11 +162,27 @@ const byGroup = (events) => {
   return sorted.reverse();
 };
```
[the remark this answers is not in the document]
mirisuzanne: `sorted.push({ group, data });` should be adding items to the array?

src/filters/events.js
```diff
@@ -107,11 +162,27 @@ const byGroup = (events) => {
...
+    default: true
+*/
+const getEvents = (collection, tag, group = true) => {
   let events = fromCollection(collection);
 
-  if (only) {
-    events = events.filter((event) => event.tags.includes(only));
+  if (tag) {
+    events = events.filter((event) => event.tags.includes(tag));
   }
```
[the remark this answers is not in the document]
mirisuzanne: This is another feature I think will very likely go away on our site. For mine, event tags can come from two places, so they have to be merged manually. That means filtering events-by-tag is an optional step here. On our site, I think we can avoid the duplication, and let you use collections to filter events by tag. 

Sorry i left in some things we aren't using yet. I should probably remove for now…

src/filters/utils.js
```diff
@@ -1,50 +1,160 @@
...
-const slugify = (text) =>
-  text
-    .toString()
-    .toLowerCase()
-    .replace(/\s+/g, '-') // Replace spaces with -
-    .replace(/[^\w-]+/g, '') // Remove all non-word chars
-    .replace(/--+/g, '-') // Replace multiple - with single -
-    .replace(/^-+/, '') // Trim - from start of text
-    .replace(/-+$/, ''); // Trim - from end of text
+  if (attrs && attrs.length) {
```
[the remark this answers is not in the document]
mirisuzanne: Above, on line 49, I re-assign it so we can pass in a string, but always count on getting an  array. Is there a better way to handle that? We're also re-assigning `obj` inside the if statement.

(I know very little about JS or best-practice here, so I need your input)

src/filters/utils.js
```diff
@@ -1,50 +1,160 @@
...
+    type: object
+  attrs:
+    type: string | array
+    note: The attributes to get
+  test:
+    type: object
+    note: attribute/value pairs to find in the resulting array
+*/
+const getJust = (data, attrs, test) => {
+  data = get(data, attrs);
```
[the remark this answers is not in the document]
mirisuzanne: yep

content/_data/filters.js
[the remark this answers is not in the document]
duggiemitchell: @mirisuzanne sure I can do that.

src/filters/utils.js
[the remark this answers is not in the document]
duggiemitchell: I think in general it's not a good practice to assign function arguments since its unclear to the caller that they'll be modified and can lead to unexpected behavior.  We should probably also include a default value for `value` if it's an optional argument.

content/_layouts/docs.njk
mirisuzanne: I think this should be fixed, if one of you wants to confirm.

src/filters/utils.js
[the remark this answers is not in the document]
mirisuzanne: ok, I can take a stab at this, but it's another place I don't quite understand JS yet, so we'll see :p

content/_data/filters.js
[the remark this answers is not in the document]
duggiemitchell: Keeping this as is since the build does not seem to serve from Browsersync using async/await

src/filters/utils.js
```diff
@@ -1,50 +1,158 @@
...
+  data:
+    type: object
+  attrs:
+    type: string | array
+    note: The attributes to get
+  test:
+    type: object
+    note: attribute/value pairs to find in the resulting array
+*/
+const getJust = (data, attrs, test) => {
```
duggiemitchell: @mirisuzanne tempted to delete this util since based on the description, it does the same as the `items` filter. 
 it's used in a couple places so left as is, so may need more info on what we'd like to accomplish with this one before refactoring.

mirisuzanne:
> with adding `lodash` as a dependency, should we be concerned about increasing our bundle size and possibly adding tree shaking to our config? (I've never used rollup before and not sure if there is a plugin needed for it to tree shake commonjs modules.)

@duggiemitchell These filters are entirely server-side. Don't know if tree-shaking is still useful, but we don't have to worry about lodash adding download for users at this point.

PR #17, opened 2020-02-07
Events

mirisuzanne:
Elephootball

- [Trello card](https://trello.com/c/Xqch4to7/6-i-want-talks-and-podcasts-and-other-events)

## Description

Filters allow you to create event-pages from any collection, either:

- adding them in place of the page (talk/workshop/etc) they represent
- adding them to the collection in addition to their origin pages
- returning only events from the collection

Event-page objects look like their source pages, but `templateContent` key is removed, and  replaced by an `event` key with all relevant event information. We also have filters to check if an event-page `isFuture` (the date or end-date has not passed), and filter a list of events to only  show the future events.

The `index` option on a page is also more powerful now, allowing you to create event indexes from any collection.

## Impacted areas of website

- Pages with events (talks & workshops) now list those events by default
- The `/services/speaking/` page includes a list of upcoming events
- The "resources" page now shows events rather than their source pages
- There is a `/sample/events/` page that demonstrates some of the options
- Updated upcoming event details on Dynamic CSS talk


mirisuzanne:
[I want talks and podcasts and other events](https://trello.com/c/Xqch4to7/6-i-want-talks-and-podcasts-and-other-events)

src/filters/pages.js
```diff
@@ -28,11 +28,13 @@ label: isCurrent
 category: Status
 note: Check that the page does not have an end date
 params:
-  collection:
-    type: array of pages
+  page:
+    type: page object
 */
```
duggiemitchell: It might be useful to include a note about the shapes of objects in the filter documentation. e.g. an example of what properties are in the `page object`.

src/filters/events.js
```diff
@@ -0,0 +1,99 @@
...
+  Build an event-page out of combined event and page data,
+  so that events can be treated as a page of their own
+params:
+  page:
+    type: page object
+  event:
+    type: object
+*/
+const buildEvent = (page, event) => ({
+  date: event ? getDate(event.date) || page.date : page.date,
```
duggiemitchell: Is `event` an optional argument?  If `buildEvent` will always be called with an `event` and we intend to check if the event has a date, it may make sense to write it like this:

```suggestion
  date: event.date ? getDate(event.date)  : page.date,
```

src/filters/pages.js
mirisuzanne: These are defined by 11ty. Maybe just saying that would be enough?

duggiemitchell: Yea, I think so :+1:

PR #18, opened 2020-02-10
Upgrade js deps.

jgerigmeyer:
gerbil


src/scss/config/_color.scss
```diff
@@ -137,9 +137,6 @@ $_ccs: (
 ///   See map for valid tokens.
 /// @return {*} -
 ///   Value from `$ccs` map, based on given key
-/// @example
-///   @use '../config/color';
-///   $prime-hue: color.ccs('prime');
```
jgerigmeyer: @mirisuzanne This was breaking Sassdoc (likewise the other SCSS examples) -- I think because https://github.com/oddbird/sassdoc-theme-herman/issues/341

src/scss/initial/_links.scss
```diff
@@ -21,14 +21,6 @@
...
 /// but the target depends on the presence or absence of icons,
 /// in order to avoid awkward-looking icon underlines.
-/// @example njk
-///   {% import 'utility.macros.njk' as utility %}
-///   <a href="#">simple text link</a>
-///   {{ utility.icon_link(
-///     icon='icons/external.svg',
-///     text='link with icon',
-///     url='#'
-///   ) }}
```
jgerigmeyer: We can't use Nunjucks examples unless we pass in a custom Nunjucks environment that has all our Eleventy filters/shortcodes added. I think we can do that if we want -- it'll require some changes to how we're calling SassDoc/Herman.

package.json
```diff
@@ -41,45 +41,47 @@
...
     "build": "run-s build:assets build:site eleventy"
   },
+  "dependencies": {},
   "devDependencies": {
-    "@11ty/eleventy": "^0.9.0",
-    "@11ty/eleventy-plugin-syntaxhighlight": "^2.0.3",
-    "@babel/core": "^7.6.4",
-    "@babel/preset-env": "^7.6.3",
-    "accoutrement": "^2.1.2",
+    "@11ty/eleventy": "^0.10.0",
```
jgerigmeyer: @mirisuzanne It's possible this upgrade includes some features we might want to use? Not sure, but take a look at: https://github.com/11ty/eleventy/releases/tag/v0.10.0

src/scss/initial/_links.scss
mirisuzanne: I think we should for-sure allow adding these docs – the question is if we need to parse & display the results. For now I'd do whichever is simpler:

- Either allow njk examples as plain text without parsing/rendering results. Basically just make it possible to silence the error
- Or pass in the njk env to parse/render results

package.json
mirisuzanne: The only update that's particularly interesting to me is converting our global/folder `json` data files to `yaml` for consistency with file data: `eleventyConfig.addDataExtension("yaml", contents => yaml.safeLoad(contents));` Not sure if it's that simple, or not. We can either play with it in this branch, or make a story to  look at it later.

jgerigmeyer: Added a story: https://trello.com/c/vhNKI3YC

jgerigmeyer:
Closing; superseded by #19

PR #20, opened 2020-02-12
JS Tests

duggiemitchell:
todo: 
- [x] tests for filters/type.js
- [x] tests for filters/time.js
- [x] fix failing tests
- [x] possible refactor of filters

## Cute animal pic
honeybadger

https://trello.com/c/fer5NY7n

## Description
_The commit messages say what you did; this should explain why and how._
- [ ] Description is in Trello card

## Steps to test/reproduce
run `yarn test` to run tests in `./test` folder

## Reviewer tasks:

- [ ] Reviewed code


package.json
```diff
@@ -35,9 +35,11 @@
     "serve": "npm-run-all build:assets build:site -p eleventy:serve watch:js watch:sass",
     "build:site": "run-s site:clean sassdoc",
     "build:assets": "npm-run-all assets:clean -p js sass",
-    "build": "run-s build:assets build:site eleventy"
+    "build": "run-s build:assets build:site eleventy",
+    "test": "jest"
   },
   "devDependencies": {
+    "#": "link:./src/filters",
```
jgerigmeyer: @duggiemitchell @mirisuzanne I added this link so we can import or require from e.g. `#/pages` instead of using the full path.

mirisuzanne: Why `#` rather than  `~`?

jgerigmeyer: Because it's specifically a shortcut to the 11ty filters, not to node modules -- I didn't want to confuse the two. 🤷‍♂ 

This is all arbitrary, of course. We could just use a generic shortcut to `src/` and do e.g. `require('#/filters/pages')` if we want.

duggiemitchell:
@jgerigmeyer or @wlonk 80% I could use some review on a couple problematic tests that are either failing or doesn't have coverage for.  I'm also not 100% sure on the best way to mock functions from outside libraries such as in the `/type.js` filters.

src/filters/pages.js
```diff
@@ -291,8 +291,7 @@ const pageType = (tags) => {
     'Links',
     'News',
   ];
-
-  return tags ? tags.find((tag) => types.includes(tag)) : false;
+  return tags.filter((tag) => types.includes(tag));
```
jgerigmeyer: This change is causing the build to fail, because 1) it is expecting a `string` return value instead of an array, and 2) `tags` might be missing entirely.

```suggestion
  return tags ? tags.find((tag) => types.includes(tag)) : false;
```

src/filters/tags.js
```diff
@@ -43,9 +42,9 @@ params:
 */
 const displayName = (tag) => {
   const capitalize = ([first, ...rest]) =>
-    first ? first.toUpperCase() + rest.join('') : '';
+    first && first.toUpperCase() + rest.join('');
```
jgerigmeyer: Won't this change mean we now get `false` instead of `''` if tag is empty?

test/js/pages.test.js
```diff
@@ -0,0 +1,165 @@
...
+  });
+
+  test('getPublic', () => {
+    const publicPages = [collection[0], collection[2]];
+
+    expect(getPublic(collection)).toEqual(publicPages);
+  });
+
+  test('hasData', () => {
+    // this keeps returning undefined...
```
jgerigmeyer: These are returning `undefined` because lodash is mocked out (I'm not sure why?). I think we should remove `jest.mock('lodash');` above.

test/js/pages.test.js
```diff
@@ -0,0 +1,165 @@
...
+  });
+
+  test('pageYears', () => {
+    const testPage = pageYears(collection)[0];
+    const pageKeys = Object.keys(testPage);
+
+    expect(pageKeys).toHaveProperty('sort');
+    // This is somehow failing but its in keys when printed //
+    expect(pageKeys).toHaveProperty('year');
+  });
```
jgerigmeyer: You're already getting the list of keys with `Object.keys()`, so these should either be checking `expect(pageKeys).toContain('sort')` or `expect(testPage).toHaveProperty('sort')`.

test/js/birds.test.js
```diff
@@ -0,0 +1,87 @@
...
+  test('returns pages from inactive birds', () => {
+    const expected = [collection[3]];
+
+    expect(active(collection, false)).toEqual(expected);
+  });
+});
+
+describe('authorPage', () => {
+  test('returns author home page', () => {
+    authorPage(collection, 'erica'); // / this is returning undefined which is not expected
```
jgerigmeyer: `authorPage` looks for a `data.bird` key on the page; the sample `collection` above does not have that.

test/js/pages.test.js
```diff
@@ -0,0 +1,165 @@
...
+    });
+    test('returns pages grouped by year', () => {
+      byYear(collection);
+    });
+  });
+
+  describe('pageType', () => {
+    test('Return one of several resource "types" from page tags', () => {
+      const tags = ['Workshops', 'Podcasts', 'foo', 'bar'];
+      expect(pageType(tags)).toEqual(['Workshops', 'Podcasts']);
```
jgerigmeyer: This should just match the _first_ occurrence, not all matching ones.

test/js/tags.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+  test('publicTags', () => {
+    const expected = [tags[0], tags[2]];
+
+    expect(publicTags(tags)).toEqual(expected);
+  });
+
+  test('tagData', () => {
+    // @todo needs a fix...
+    tagData(collection, tags, 'pageCount');
```
jgerigmeyer: The first arg here is expected to be an object with all collections by tag.

jgerigmeyer:
> I'm also not 100% sure on the best way to mock functions from outside libraries such as in the `/type.js` filters.

@duggiemitchell Hm, that's a good question. Since those are relatively simple libraries, I'd probably start by _not_ mocking them at all. Then our tests are more similar to reality, even if we're also testing that those libraries do what we think they do.

src/filters/events.js
```diff
@@ -76,8 +76,8 @@ params:
   page:
     type: event object
 */
-const isFuture = (event) =>
-  event.end ? getDate(event.end) >= now : getDate(event.date) >= now;
+const isFuture = (page) =>
+  page.event ? getDate(page.event.end) >= now : getDate(page.date) >= now;
```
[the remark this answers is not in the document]
duggiemitchell: Yes, I believe so.  :+1:  can change it back to check for events...

src/filters/events.js
```diff
@@ -98,6 +98,7 @@ params:
     type: array
     note: containing 11ty page objects
 */
+/* istanbul ignore next */
```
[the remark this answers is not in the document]
duggiemitchell: @mirisuzanne talked about changing how we filter events and ignoring this filter since it's not being used anywhere atm.

src/filters/pages.js
```diff
@@ -193,13 +193,13 @@ params:
...
+  /* will pages ever not have data attributes? */
+  /* istanbul ignore if */
   if (!page.data) {
     return undefined;
   }
 
-  return page.data.renderData
-    ? page.data.renderData[key] || page.data[key]
-    : page.data[key];
+  return page.data.renderData ? page.data.renderData[key] : page.data[key];
```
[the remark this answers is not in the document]
duggiemitchell: @jgerigmeyer Makes sense, think I was reading this as either if `renderData` exists, return the page with matching key or the page.  Could have also been trying to get a test to pass and seems to work fine now after reverting the change.

test/js/pages.test.js
duggiemitchell: OK makes sense.  In `pageType` I feel like it would make more sense to return an empty string instead of false if the tag isn't included in the types list.

src/filters/pages.js
```diff
@@ -193,13 +193,13 @@ params:
     type: string
 */
 const render = (page, key) => {
+  /* will pages ever not have data attributes? */
```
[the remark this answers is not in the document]
jgerigmeyer: Ah, found it. https://www.11ty.dev/docs/collections/#collection-item-data-structure

So I guess pages _within a collection_ have stuff on `page.data`, but pages outside of a collection have their stuff on just `page`?

I have some concerns about the maintainability of the current structure, with a mix of Nunjucks and Markdown, lots and lots of JS filters and lots of Nunjucks macros. There are so many small pieces (which can be a plus) -- in practice it's very difficult for me to understand what any one piece is actually doing. It feels like lots of rabbit holes.

I'll try to think a bit more about what precisely is the issue I'm trying to articulate, and what I'd recommend to resolve it.

src/filters/pages.js
```diff
@@ -254,12 +254,12 @@ params:
     note: optionally include events in sort/year dates
 */
 const byYear = (collection, events = true) => {
+  /* istanbul ignore else */
```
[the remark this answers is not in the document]
duggiemitchell: No reason I can think of now, removing...

src/filters/pages.js
duggiemitchell: OK.  Does it make sense for this to return an empty string instead of false if the tag isn't included in the list? (I think I may have made that change and the build is successful.)

jgerigmeyer: Hm. I'm not sure practically, but in some ways `false` makes more sense to me theoretically. Or `undefined`?

src/filters/tags.js
duggiemitchell: Yup you are right.  Maybe I'm misunderstanding how this is expected to called; can't seem to get the coverage for returning an empty string :woman_shrugging:

src/filters/pages.js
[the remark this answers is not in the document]
mirisuzanne: Yeah, I find the difference between the two totally frustrating. I agree there's a lot going on. I'm not sure I know the fix…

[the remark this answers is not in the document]
duggiemitchell: I will say it's not always clear to me what the data looks like (for things like pages and collections) and that sort of makes me feel the testing is brittle (if that makes sense). I may be worse at articulating what exactly my frustration is, so  I am interesting to hear you all's thoughts on how things are structured.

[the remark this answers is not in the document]
mirisuzanne: There are two page objects, the link Jonny shred (a collection-page) and the current [`page` variable](https://www.11ty.dev/docs/data-eleventy-supplied/#page-variable-contents). They have a lot of overlap (`url`, `date`, `inputPath`, `outputPath`, `fileSlug`), but a few differences:

- collection-pages have `templateContent` and `data` (which includes page, directory, and global data)
- current `page` has `filePathStem`, but the content and data-properties are not included. We have to access them un-scoped as `content`, and `<property>`.

I find it very tempting to do `set page = collections.all | getPage(page.url)` first thing in our layouts, so that the page object includes all data and content. We would just have to make sure that layout is not used by any page excluded from collections. (or maybe there's no reason to exclude pages from collections…)

duggiemitchell:
@jgerigmeyer I think I've addressed the review comments and have gone as far as I can without direction on a couple tests (especially related to `typopgraphy filters`) and I did make an external file to import collections, that may use some cleanup? (as in not representative of how collections actually look).

test/js/birds.test.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+  test('returns "oddbird" authored pages', () => {
+    const expected = [collection2[1], collection2[2]];
+
+    expect(getPages(collection2, 'erica')).toEqual(expected);
+  });
+});
+
+describe('active', () => {
+  test('filters pages from inactive birds', () => {
```
mirisuzanne: maybe clarify that these are not bird-authored pages, but bird-detail pages?

```suggestion
  test('filters inactive bird-detail pages', () => {
```

test/js/birds.test.js
```diff
@@ -0,0 +1,35 @@
...
+    const expected = collection2[3];
+
+    expect(active(collection2)).toEqual(expect.not.objectContaining(expected));
+  });
+  test('returns pages from inactive birds', () => {
```
mirisuzanne:
```suggestion
  test('returns inactive bird-detail pages', () => {
```

jgerigmeyer:
There's more cleanup I'd like to do of course, but I'm good with this PR once @duggiemitchell and @mirisuzanne sign off 👍

jgerigmeyer:
[I want JS tests for back-end (eg filters) JS](https://trello.com/c/fer5NY7n/32-i-want-js-tests-for-back-end-eg-filters-js)

PR #22, opened 2020-02-13
[resources] Resource lists

mirisuzanne:
squirrel!

- [Trello card](https://trello.com/c/qlXEAjrf/33-what-does-the-resources-list-look-like-are-there-different-styles-for-different-types-of-resources)
- [Preview](https://deploy-preview-22--oddleventy.netlify.com/)

## Description
Generating and styling resource lists. 

## Impacted areas of website

The primary resource list is at `/blog/` (the top-nav "Resources" link), but resource lists also appear on: 
- generated tag pages
- any page with `index` set in yaml data
- bird author pages

I also tried to re-use logic where I could, so all page headers now rely on some of the same macros that resource-lists use.


.gitignore
```diff
@@ -1,7 +1,7 @@
 _site
 _built
 
-TAGS
+# TAGS
```
mirisuzanne: @jgerigmeyer this came from a yarn/rollup commit, but it ends up ignoring our tag pages. Feel free to delete or replace with something more targeted?

content/_includes/devices/iphone.svg
```diff
@@ -1,13 +1,12 @@
...
 	data-device="iphone"
-  aria-labelledby="iphoneID"
   role="presentation"
   xmlns="http://www.w3.org/2000/svg"
   width="250px"
   height="500px"
 	viewBox="0 0 250 500"
 >
-<title id="iphoneID">iphone</title>
+<title>iphone</title>
```
mirisuzanne: Since icons are repeated, we were causing duplicate-ID issues

content/_includes/embed.macros.njk
```diff
@@ -293,11 +294,7 @@ params:
 ) -%}
 {%- if src -%}
 {%- set src = src if ('://' in src) else ('/assets/images/' + src) -%}
-{%- set main = {
-  'src': src,
-  'alt': alt
-} -%}
-<img {{ utility.show_attrs(main) }} {{ utility.show_attrs(attrs) }}/>
+<img src="{{ src }}" alt="{{ alt }}" {{ utility.show_attrs(attrs) }}/>
```
mirisuzanne: Allways an alt attr, even when empty

.gitignore
jgerigmeyer: This is a holdover from old projects, not added intentionally. Just remove it.

mirisuzanne:
[What does the Resources List look like? Are there different styles for different types of Resources?](https://trello.com/c/qlXEAjrf/33-what-does-the-resources-list-look-like-are-there-different-styles-for-different-types-of-resources)

content/_includes/devices/iphone.svg
stacyk: and I believe this title tag, when used in a link, is read via screen reader. So right now, on a post on the resrouce page, it reads that the item I am on is a Link, The SVG title= "link' for instance on the Link Posts, and the Span text, which in that example is "link". So, it says "Link link link" for each of the link icons.

stacyk: If possible, maybe the span text there should be view all posts tagged with "link" or something of the like.

stacyk: Screen Shot 2020-02-18 at 4 21 37 PM
Screen Shot 2020-02-18 at 4 20 29 PM

stacyk: Screen Shot 2020-02-18 at 4 19 15 PM

stacyk:
Oh, forgot to mention, a screen reader also reads out the "space" when widont kicks in. I wonder if the extra code and this screen reader issue is worth the little design boost we get?
 
Screen Shot 2020-02-18 at 4 28 59 PM

mirisuzanne:
@stacyk block-level links were a feature of HTML5. I don't know about accessibility - don't see any recent articles with a quick search. I know AEA uses them a lot (I think Eric was the one initially pushing for the feature). If a labeled-by attribute would help, we can do something like that.

content/_includes/devices/iphone.svg
mirisuzanne: Asked around a bit, and it sounds like screen readers are inconsistent with inline SVG, and the recommendation is to use `aria-hidden focusable="false"` on the icons, no need for `<title>`, and label them in HTML. I think your proposal for a bit more label context is also good.

src/scss/components/_post.scss
```diff
@@ -0,0 +1,186 @@
...
+  --hero-border: 1px;
+  clip-path: var(--crop);
+}
+
+[data-hero='card'] {
+  --hero-height: calc(var(--gutter) * 8);
+  margin: calc(var(--gutter) * -1) calc(var(--gutter) * -1) var(--newline);
+}
+
+[data-media='hero'] {
```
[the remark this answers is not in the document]
mirisuzanne: angled headers were overlapping the nav. fixing…

mirisuzanne:
@jgerigmeyer builds are failing with: `Usage Error: Couldn't find a script named "prod".`

jgerigmeyer:
> @jgerigmeyer builds are failing with: `Usage Error: Couldn't find a script named "prod".`

Yeah, Netlify doesn't allow different settings per-branch, so it's setup for `build-oss-docs` to build right now.

PR #23, opened 2020-02-14
Upgrade js deps and node.

jgerigmeyer:
Junebug19


scripts/compile-sass.js
```diff
@@ -62,16 +62,17 @@ const nodeImporter = (url, prev, done) => {
   }
 };
 
-const outputFile = (file, contents) =>
+const outputFile = (file, inputName, contents) =>
   fs.outputFile(file, contents.toString().trim(), (writeErr) => {
-    console.log(`Compiled: ${file}`);
+    console.log('\x1b[32m%s\x1b[0m', `Compiled Sass: ${inputName} => ${file}`);
```
wlonk: I would advocate for making this legible by pulling the color codes into well-named variables and using those by name here, but that's not so important that it'll block my approval on this PR

PR #27, opened 2020-02-21
Basic page template using grids

mirisuzanne:
hair

- [Trello card](https://trello.com/c/bjo5Xp0u/12-i-want-structured-page-templates)
- [deploy preview](deploy-preview-27--oddleventy.netlify.com/)

## Description
- Removed `page-margin` from the body, and applied it as-needed for each section
- Set up `data-layout='main'` as a grid with several named column-areas:
   - `page` (general fallback for content)
   - `wide` (used by figures, codepens, `.extend-small`, etc)
   - `content` (used by `data-grid` blocks and `.extend-large`, retains page margins)
   - `full` (used by blockquotes, section titles, anything with bleed, and `.extend-full`)
- I also set up two lodash-based filters for concat/merge of arrays/objects in nunjucks, and use one of them. I think there would be more places to use both, as a cleanup step at some point.

## Impacted areas of website
…basically every page. But if you look at some of the pages touched by this commit, you'll get a good sense.


mirisuzanne:
[I want structured page templates](https://trello.com/c/bjo5Xp0u/12-i-want-structured-page-templates)

src/scss/patterns/_sizes.scss
```diff
@@ -16,22 +12,22 @@
 
 // Extend thing outside its container on one side only
 .extend-left {
-  @include extend-side('left');
+  --column: wide / page;
```
stacyk: I don't think I've seen this `/` syntax used anywhere. Or is this literally dividing these two values? lol

mirisuzanne:
@stacyk I think I addressed the issues here, but take a look. For now I wrapped floats, but we can create a new story to look at images and address remaining issues. There are also a lot of images hard-coded, which we can cleanup. I think that's a new task – this branch just needs to provide a consistent template we can work in.

src/scss/patterns/_sizes.scss
mirisuzanne: This is being used in the `grid-column` property, where `/` separates `grid-column-start` and `grid-column-end`. Combined with area/line names, `wide / page` starts at the `wide-start` line, and ends at the `page-end` line.

stacyk: oh, yeah, ha ha. That makes sense.

PR #29, opened 2020-02-24
Valid RSS feed

jgerigmeyer:
ToDo:

- [x] valid RSS feed
- [x] remove OSS and client work from feed
- [x] update dates to use UTC time zone?

## Cute animal pic

hare

https://trello.com/c/vBQCWXX3

---

## Reviewer tasks:

- [ ] Reviewed code


mirisuzanne:
[I want a validated RSS/Atom feed](https://trello.com/c/vBQCWXX3/49-i-want-a-validated-rss-atom-feed)

content/_includes/page/list.njk
```diff
@@ -17,6 +17,8 @@
 {% elif bird and (bird != 'oddbird') %}
   {% set list = collections._post | byBird(bird) %}
   {{ post.by_year(list, collections) }}
+  {% set oss = collections["Open Source"] | byBird(bird) %}
+  {{ post.by_year(oss, collections, 'Open Source', Infinity) }}
```
jgerigmeyer: @mirisuzanne So. I removed OSS and Client Work (without a Case Study) from the `_post` collection. I _think_ that removes those from the Resources list, and also from the RSS feed.

But... from what I can tell, there were two other places we were using the `_post` collection:

1. Deciding who from the birds-emeritus has a page to link to. I decided contributing to OSS doesn't really need to require a standalone page. Practically, this makes no difference at the moment.
2. Listing Resources on an individual Bird page. I made some changes so that it list "Open Source" below "Resources"... but I'm not tied to it.

jgerigmeyer:
@mirisuzanne Still one or two small changes I want to make, but happy for your review/cleanup any time.

content/_includes/page/list.njk
mirisuzanne: I think Sondra has a different plan for OSS list layout, but this is a good placeholder. I moved the `title` arg to the `list()` macro, so we can save `by_year` to do the yearly grouping logic. No need to call it if we're using `Infinity` to avoid it's core feature.

content/blog/2019/smashing.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Smashing CSS Follow-Up Q&nbsp;&&nbsp;A
+title: Smashing CSS Follow-Up Q&A
```
mirisuzanne: Didn't like my hack for getting a fancy ampersand?

jgerigmeyer: I mean, I liked the result, but it does feel like a hack, and we're not using it consistently. Elsewhere we use `Q&A`. I guess I'd vote for a more robust and consistent solution?

mirisuzanne: I agree. Doesn't seem high priority, tho.

jgerigmeyer: Yeah, agreed. The reason I discovered it is because the RSS validator complained about the HTML escaped entity in the `<title>` attr. Right now we assume titles are text, and only escape the contents.

PR #30, opened 2020-02-24
Image brightness, and aria landmarks

mirisuzanne:
bioluminescence

- [Trello card](https://trello.com/c/eOUyrp3T/35-i-want-images-to-be-less-intense-in-dark-mode-or-maybe-based-on-lightness-contrast) (partial)
- [Preview](https://deploy-preview-30--oddleventy.netlify.com/)

## Description
This branch went a bit beyond the original Trello card, because testing revealed a few other issues.

- I added the palette icon for the color menu
- Images now respond to a combination of light/dark mode, lightness, and contrast set in CCS.
- On hover/focus, they return to neutral state (`brightness(1) contrast(1)`)
- I was running into aria-landmark issues when I did accessibility testing, so I also re-arranged and labeled some landmarks on the site template. Not strictly color related…

## Steps to test/reproduce
- go to a page with a header image (see the first resource listed), and play with color mode/lightness/contrast settings
- navigate the site with a screen reader


mirisuzanne:
[I want images to be less intense in dark mode, or maybe based on lightness/contrast?](https://trello.com/c/eOUyrp3T/35-i-want-images-to-be-less-intense-in-dark-mode-or-maybe-based-on-lightness-contrast)

src/scss/initial/_colors.scss
```diff
@@ -26,4 +26,9 @@
   @include tools.tokens--(config.$ccs-colors);
   @include tools.tokens--(config.$background-colors);
   @include tools.tokens--(config.$content-colors);
+
+  --img-bright--base: #{ccs.if-mode($light: 40%, $dark: 30%)};
```
stacyk: it's cool that we can do this.

mirisuzanne:
@stacyk - if you go to a page with a (non-svg) hero image, and play with lightness/contrast in the color menu, you can see how images now respond to those settings.

I pushed an update so they only return to default lightness/contrast on hover *inside links*. Take a look and see what you think.

PR #46, opened 2020-03-31
Image minification

duggiemitchell:
## Cute animal pic
little-big-dog

## Links
[Trello card](https://trello.com/c/EamUpG9W)

## Description
- All (I think) images should be resized and have `srcset` attributes

## Steps to test/reproduce
_Please explain how to best reproduce the issue and/or test the changes locally._

## Show me
Screenshot from 2020-03-31 18-31-54

---

## Reviewer tasks:

- [ ] Reviewed code


mirisuzanne:
[I want automated responsive images & thumbnails](https://trello.com/c/EamUpG9W/44-i-want-automated-responsive-images-thumbnails)

content/_includes/embed.macros.njk
```diff
@@ -323,7 +323,13 @@ params:
 ) -%}
 {%- if src -%}
 {%- set src = src if ('://' in src) else ('/assets/images/' + src) -%}
-<img src="{{ src }}" alt="{{ alt }}" {{ utility.show_attrs(attrs) }}/>
+<img
+  srcset="{{src | imgSuffix('320')}} 320w, {{src | imgSuffix('550')}} 550w"
+  sizes= "(max-width: 550px) calc(100vw - 120px), 550px"
+  src="{{ src }}"
```
mirisuzanne, line 329: In the MDN examples, they duplicate the default image in both places. I'm going to look into that a bit, and see if it's necessary… 

I also wonder if we want to define our largest size as well (assuming many images will be larger than `550` by default), maybe something like `1200`?

mirisuzanne: ok, Eric Portis pointed me to [the spec](https://html.spec.whatwg.org/multipage/images.html#introduction-3) which says that `src` is ignored when using `w`. So I think we do need to represent larger images in the `srcset` somehow.

duggiemitchell: I think using `src` is helpful for unsupported browsers. Creating default image size at 1200 make sense.  I'm (doubtful, but) curious if this affects images that are naturally smaller. Would it load images that are too big for what we need?

src/filters/type.js
```diff
@@ -8,14 +8,40 @@ const removeMd = require('remove-markdown');
...
+        {
+          width: 550,
+          rename: {
+            suffix: '-550',
+          },
+        },
+      ],
+    },
+    sizes: {
+      '*': '(max-width: 550px) calc(100vw - 120px), 550px',
```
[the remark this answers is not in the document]
duggiemitchell: Sure, I can play with that, unless we decide leaving out `sizes` altogether and letting the browser make that choice instead.

content/_includes/embed.macros.njk
mirisuzanne: - Yeah, we don't want to remove `src`, but we do need to duplicate that path in the `srcset`
- If possible, I think we should only shrink images, and never enlarge them.

gulpfile.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+const gulp = require('gulp');
+const responsive = require('gulp-responsive');
```
jgerigmeyer, line 3: I'm open to this if it's the only way we can make it work... but I'm hesitant to add `gulp` unless we have to. I'm wondering if we could use something like https://github.com/MarcusCemes/responsive-image-builder to do the same thing? I'd have to dig into it a bit more.

duggiemitchell:
@mirisuzanne I think I made the right fixes for your re-review. 

Note that I found something weird with the markdown plugin that we are using; it errs if there is no key for sizes (my assumptions were omissions are OK), so that key is left empty in our config.

gulpfile.js
```diff
@@ -0,0 +1,42 @@
...
+
+gulp.task('images', () =>
+  gulp
+    .src(`src/images/**/*.{jpg,png}`)
+    .pipe(
+      responsive(
+        {
+          // Resize all JPG images to three different sizes: 320, 550, and 1200 pixels
+          '**/*': [
```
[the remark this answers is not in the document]
duggiemitchell: oops, @mirisuzanne looks like you took care of this here: 742aa741b7d8702587e4bd16080607cbec73841a

Is this good to go, or are there other changes needed around this (other than fixing tests)?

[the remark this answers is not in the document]
mirisuzanne: @duggiemitchell I took care of it in the js filters, but not in the gulpfile. It was being hard-coded in both places. I wasn't sure if the same solution would work for the gulpfile. Can you take a look at that one still?

gulpfile.js
duggiemitchell: I vote for merging this and seeking alternatives in another card for later; but could defer to @jgerigmeyer 's thoughts on that.

content/_includes/embed.macros.njk
```diff
@@ -320,18 +341,34 @@ params:
...
   src,
   alt='',
+  width=none,
+  height=none,
   attrs={}
 ) -%}
 {%- if src -%}
 {%- set src = src if ('://' in src) else ('/assets/images/' + src) -%}
-<img src="{{ src }}" alt="{{ alt }}" {{ utility.show_attrs(attrs) }}/>
+{% set set = srcset(src, width) %}
```
jgerigmeyer, line 363: whoa -- setting a variable to the name "set" seems like asking for trouble 🙈

content/_includes/embed.macros.njk
```diff
@@ -309,6 +315,21 @@ params:
 {%- endif -%}
 {%- endmacro %}
 
+{% macro srcset(
+  src,
+  width=none
+) -%}
+  {%- set width = width or 640 -%}
```
jgerigmeyer, line 322: @mirisuzanne Any reason to do this vs just setting `width=640` above?

content/_includes/embed.macros.njk
```diff
@@ -309,6 +315,21 @@ params:
...
+  {%- set sizes = 'srcset' | fromTaxonomy -%}
+  {%- if width > sizes[0] -%}
+    {%- for size in sizes -%}
+      {%- if size < width -%}
+        {{- src | imgSuffix(size | string) }} {{ size + 'w,' -}}
+      {%- endif -%}
+    {%- endfor -%}
+    {{- src }} {{ width + 'w' -}}
+  {%- endif -%}
+{%- endmacro %}
```
jgerigmeyer, line 332: @mirisuzanne Is it intentional that we're only using this in `figure()` and `post.hero()` images?

content/_includes/embed.macros.njk
mirisuzanne: I'm not actually sure… Do defaults work if the calling location passes `none` as the value?

jgerigmeyer: Yes, the default value is used in place of `none`.

content/_includes/embed.macros.njk
mirisuzanne: I don't know what you mean. `post.hero()` does call `embed.img()`, which applies `srcset()`. I'm seeing srcsets on all relevant images when I look at the deploy preview. What am I missing?

mirisuzanne: We use `embed.img` all over the site. Do you just mean that not all images have the height/width set? I figured this was worth merging before we take the time to update every single image.

mirisuzanne: I've been adding dimension to every image I come across as I work, but we'll need a card to do it more intentionally

src/filters/type.js
```diff
@@ -12,12 +12,14 @@ const markdownItResponsive = require('@gerhobbelt/markdown-it-responsive');
...
-  },
-}));
+const responsiveConfig = taxonomy.srcset
+  .map((size) => ({
+    width: size,
+    rename: {
+      suffix: `-${size}`,
+    },
+  }))
+  .concat([{}]);
```
jgerigmeyer, line 22: @mirisuzanne @duggiemitchell This empty object means that the fallback original image is also processed and passed through along with the new resized ones.

mirisuzanne: That's great, but without supplying the width, it results in[ invalid `srcset`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-srcset):

- "If no descriptor is specified, the source is assigned the default descriptor of 1x."
- "It is incorrect to mix width descriptors and pixel density descriptors in the same srcset attribute."

I'd love to talk through some of the isseues here, around supplying a width to markdown. Or we can try to move away from md images entirely to using `embed.img()` macro.

jgerigmeyer: I don't follow... All this does is process the default images through the optimizer as well -- this change shouldn't impact the `srcset` attr at all.

mirisuzanne: well maybe it's a problem with the plugin then? The result of a md image is not currently valid:

```html
<img src="/assets/images/blog/2016/cssconf/sd.jpg" alt="Sarah Drasner speaking" srcset="/assets/images/blog/2016/cssconf/sd-480.jpg 480w, /assets/images/blog/2016/cssconf/sd-960.jpg 960w, /assets/images/blog/2016/cssconf/sd-1280.jpg 1280w, /assets/images/blog/2016/cssconf/sd-2240.jpg 2240w, /assets/images/blog/2016/cssconf/sd.jpg ">
```

content/_includes/embed.macros.njk
mirisuzanne: It looks like you changed this, and changed it back?

jgerigmeyer: Yeah, it works with `none`, but apparently not with `""`. 🤷

content/_data/taxonomy.yaml
```diff
@@ -27,3 +27,17 @@ oss:
...
     show: contributor
+img:
+  # The size to use in src attribute, when srcset is not supported
+  fallback: 1280
+  # Sizes to generate automatically
+  srcset:
+    - 480
+    - 960
+    - 1280
+    - 2240
```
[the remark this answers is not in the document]
mirisuzanne: I figured our industry has a lot of large "4k" or at least `2x` density screens, where `2240`  would be used for `1120px` width?

PR #51, opened 2020-04-27
Resize image height/width attrs, and fix typos

mirisuzanne:
## Cute animal pic

mouse in a shoe

## Description
_The commit messages say what you did; this should explain why and how._
- Fixed some typos
- Image height/width are now scaled down to match fallback-image size (1280×???)

### DIFF
--- a/content/_includes/embed.macros.njk
+++ b/content/_includes/embed.macros.njk
@@ -387,7 +387,7 @@ params:
   {%- set respSrc = src | imgSuffix(fallback or config.fallback) -%}
   {%- set configSizes = config.sizes[sizes].html if config.sizes[sizes] else none -%}
 {%- endif -%}
-{% set resize = width | imgSize(height, sizes) %}
+{%- set resize = width | imgSize(height, sizes) -%}
 {%- set attrs = attrs | merge({
   'src': respSrc or src,
   'alt': alt,
--- a/content/_includes/embed.macros.njk
+++ b/content/_includes/embed.macros.njk
@@ -319,7 +319,7 @@ params:
 
 {# @docs
 label: srcset
-cateogry: Embed
+category: Embed
 note: Generate a `srcset` attribute for responsive images
 params:
   src:
@@ -385,12 +385,13 @@ params:
   {%- set imgSrcSet = imgSrcSet if (imgSrcSet != "") else none -%}
   {%- set respSrc = src | imgSuffix(config.fallback) -%}
 {%- endif -%}
+{% set resize = width | imgSize(height) %}
 {%- set attrs = attrs | merge({
   'src': respSrc or src,
   'alt': alt,
   'srcset': imgSrcSet,
-  'width': width,
-  'height': height,
+  'width': resize.width,
+  'height': resize.height,
   'sizes': (config.sizes[sizes] or sizes) if imgSrcSet else none
 }) -%}
 <img {{ utility.show_attrs(attrs) }}/>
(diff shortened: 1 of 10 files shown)


mirisuzanne:
@stacyk @jgerigmeyer the main feature here is in response to Stacy's comment about the home page image having height/width attributes set much larger than the image will ever actually display. 

- I think the simplest way to supply width/height manually is by using the actual image dimensions. Sometimes those numbers are much larger than we're actually referencing in image `src` or `srcset`.
- In most cases, browsers are only using these attributes in order to get the image ratio right while loading, so it's not important for the numbers to exactly match the output. But it's possible for the numbers to be used if CSS fails to load for any reason.
- There's no way for us to match exactly what's coming from the `srcset`, since that's dynamic, but we can more accurately reflect the fallback image `src` size. That's what this PR does - using either the fallback for a given type of image, or the default fallback of `1280`.

content/_includes/embed.macros.njk
```diff
@@ -380,18 +380,21 @@ params:
...
 {%- set attrs = attrs | merge({
   'src': respSrc or src,
   'alt': alt,
   'srcset': imgSrcSet,
-  'width': width,
-  'height': height,
-  'sizes': (config.sizes[sizes] or sizes) if imgSrcSet else none
+  'width': resize.width,
+  'height': resize.height,
+  'sizes': configSizes or sizes
```
jgerigmeyer, line 397: `sizes` is a string here -- is it (e.g. `img sizes="foobar"`) a meaningful fallback, or should we just use use `configSizes` or `none`?

content/_includes/embed.macros.njk
```diff
@@ -380,18 +380,21 @@ params:
...
 {%- if not ('://' in src) -%}
+  {%- set fallback = config.sizes[sizes].fallback if config.sizes[sizes] else none -%}
   {%- set src = ('/assets/images/' + src) -%}
   {%- set imgSrcSet = srcset(src, width) -%}
   {%- set imgSrcSet = imgSrcSet if (imgSrcSet != "") else none -%}
-  {%- set respSrc = src | imgSuffix(config.fallback) -%}
+  {%- set respSrc = src | imgSuffix(fallback or config.fallback) -%}
+  {%- set configSizes = config.sizes[sizes].html if config.sizes[sizes] else none -%}
 {%- endif -%}
+{% set resize = width | imgSize(height, sizes) %}
```
jgerigmeyer, line 390:
```suggestion
{%- set resize = width | imgSize(height, sizes) -%}
```

content/_includes/embed.macros.njk
mirisuzanne: While `sizes="foobar"` isn't useful, this is to allow either arbitrary or predefined sizes: 

- `configSizes` will grab e.g. `taxonomy.img.sizes['media'].html` (which is also a string) if `sizes == 'media'`
- Fallback to `sizes` means we can explicitly pass in any one-off sizes string like `'50vw'` or `'(min-width: 30em) 50vw, 100vw'`…

mirisuzanne:
> I don't really understand what this solves or how it does it, but 👍 😂

@jgerigmeyer the quick version is two parts: 

- it updates the fallback `src` based on the image `sizes` (when predefined)
- it updates the actual `height`/`width` attributes to match the used fallback, while maintaining the proper image aspect ratio (when available).

content/index.md
```diff
@@ -106,9 +106,10 @@ along with refactors for **integrated design systems,
 accessibility, performance,
 and long-term sustainability**.
 
-Since then OddBird has become an industry leaders --
+Since then OddBird has become an industry leader --
 from our work on Django, Sass, and Susy,
-to the **Mozilla Developer Channel**,
+to the Mozilla Developer Channel,
+**CSS Working Group**,
```
stacyk, line 112: 🔥

PR #52, opened 2020-06-02
Upgrade deps

jgerigmeyer:
penguin


src/scss/config/_color.scss
```diff
@@ -140,9 +140,6 @@ $_ccs: (
 ///   See map for valid tokens.
 /// @return {*} -
 ///   Value from `$ccs` map, based on given key
-/// @example scss
```
jgerigmeyer, line 143: @mirisuzanne These are broken, I think until we find a way to pass a custom import resolver through Dart Sass via Herman.

src/scss/initial/_links.scss
```diff
@@ -21,14 +21,6 @@
 /// Text underlines are also applied,
 /// but the target depends on the presence or absence of icons,
 /// in order to avoid awkward-looking icon underlines.
-/// @example njk
```
jgerigmeyer, line 24: @mirisuzanne This was breaking because our macros use filters added via Eleventy. We need to use the same Nunjucks environment both places, probably via https://www.11ty.dev/docs/languages/nunjucks/#optional-use-your-nunjucks-environment

package.json
```diff
@@ -43,61 +43,60 @@
   },
   "devDependencies": {
     "#": "link:./src/filters",
-    "@11ty/eleventy": "^0.10.0",
+    "@11ty/eleventy": "^0.11.0",
```
jgerigmeyer, line 46: @mirisuzanne Check out the release notes: https://github.com/11ty/eleventy/releases/tag/v0.11.0

The one I notice is that `renderData` is deprecated, replaced by https://www.11ty.dev/docs/data-computed/

jgerigmeyer:
[Look into slow dev watcher.](https://trello.com/c/uda4EeJX/43-look-into-slow-dev-watcher)

src/scss/config/_color.scss
mirisuzanne: Makes sense. I wonder if we should just remove the annotation, and leave the examples in markdown prose?

mirisuzanne:
@jgerigmeyer there seems to be an issue when I save changes in Sass, they don't show up - even with a hard refresh. I have to restart the server. Is that related to the incremental builds?

jgerigmeyer:
> @jgerigmeyer there seems to be an issue when I save changes in Sass, they don't show up - even with a hard refresh. I have to restart the server. Is that related to the incremental builds?

Yep -- https://github.com/11ty/eleventy/issues/1174

Fixed.

PR #56, opened 2020-06-23
Upgrade deps

jgerigmeyer:
duck


src/scss/config/_scale.scss
```diff
@@ -17,6 +17,9 @@ $ratios: (
   'line-height': 1.4,
 );
 
+@include tools.add-ratios($ratios);
+@include herman.herman-add('ratios', 'ratios', $ratios);
+
```
jgerigmeyer, line 22: @mirisuzanne Herman was showing a warning without these.

I also notice that VS Code throws an error at the bottom of this file where we're trying to set `tools.$breakpoints`:

```
colon expectedscss(css-colonexpected)
```

mirisuzanne: I'd  bet that comes from a linter that doesn't understand Sass module syntax?

jgerigmeyer: Hm, it's VS Code itself, not a linter. I think we ran into this earlier also, specifically in cases of assigning to a namespaced variable map?

PR #59, opened 2020-06-25
[services] Design and Development Services Page

stacyk:
## Cute animal pic
paige-cody-TgbyVWS9UDY-unsplash

https://trello.com/c/bruLAag5/63-i-want-fleshed-out-content-for-the-design-and-development-service-detail-page

## Description
Checklist of progress is in Trello as well

## Steps to test/reproduce
view the  /services/development/ page and the XD File: https://xd.adobe.com/view/b8f11a7c-c6a3-4524-410f-d08c6df679b3-cec0/screen/bb6add8a-7784-40d2-b82b-e649fa02174c/Service-Tag-Page

## Show me
WIP


stacyk:
[I want fleshed out content for the Design and Development Service detail page.](https://trello.com/c/bruLAag5/63-i-want-fleshed-out-content-for-the-design-and-development-service-detail-page)

content/services/development.md
```diff
@@ -8,10 +8,151 @@ image:
...
 ---
+
+{% import 'embed.macros.njk' as embed %}
+{% import 'quotes.macros.njk' as quotes %}
+{% import 'layout.macros.njk' as layout %}
+
+{{ quotes.find(
+  collections.all,
+  slugs=['investment', 'curious']
+) }}
```
stacyk, line 71: @mirisuzanne the Emmy's Workshop quote is on a page that is excluded from collections. Is there a good way around this to show that quote on this page?

if not, @SondraE is there a different quote we could use on this page instead of the "realistic" one that's shown on page 9 of your XD file: https://xd.adobe.com/view/b8f11a7c-c6a3-4524-410f-d08c6df679b3-cec0/screen/bb6add8a-7784-40d2-b82b-e649fa02174c/Service-Tag-Page

content/services/development.md
```diff
@@ -8,10 +8,151 @@ image:
...
+
+{% endcall %}
+
+## Our Process Is ...
+
+
+{{ layout.grid(
+  name='special-bg',
+  content=process
+) }}
```
stacyk, line 156: @mirisuzanne I wonder if I should be using teh offset grid class on this layout.grid item, but if I do, then the inner contents also need offset-item applied and I don't think I am able to do that from this .md page?

mirisuzanne: I don't think we need an offset-grid here? We use that for layouts like the work samples page, and for multiple quotes. This grid looks great as-is, except for some cleanup:

- needs padding (`var(--spacer) var(--page-margin)`?) on the special-background container
- needs the headline inside the container, which might mean…
- we want to separate the special-background container from the process grid? (Maybe also give it a more meaningful name?)

Overall, I think this is coming together nicely. But I'm happy to talk through any other questions you have.

content/services/development.md
```diff
@@ -9,9 +9,180 @@ action:
...
+      height=1611,
+      sizes='media'
+    )
+  )
+) %}
+
+{% call layout.grid('narrow-columns') %}
+
+{% call layout.block('item') %}
+
```
stacyk, line 43: @mirisuzanne @jgerigmeyer I wasn't sure if line break matter as much here as they were in RST blog. I am not sure if we have a preferred amount of space between these calls and the end ones (there are three in a row after the second column of content).

mirisuzanne: @stacyk Generally, spacing shouldn't matter too much – but it can have some impact on how the file is read as markdown. We don't have specific conventions, I tend to base it on what blocks are related or what seems most readable.

src/scss/patterns/_lists.scss
```diff
@@ -0,0 +1,58 @@
...
+/// Apply spacing after lists.
+/// @group type
+ul,
+ol {
+  margin-bottom: var(--newline);
+  margin-top: 0;
+}
+
+ul,
+ol {
```
mirisuzanne, line 16: duplicate selectors (throws a linting error)

content/_includes/layout.macros.njk
```diff
@@ -21,11 +20,8 @@ params:
       but the CSS defaults to a centered layout
       if no valid alignment is given.
 #}
-{% macro title(
-  content,
-  align='right'
-) %}
-<h2 data-section-title="{{ align }}">{{ content | mdInline | safe }}</h2>
+{% macro title(content, align = 'right') %}
```
mirisuzanne, line 23: Do you have a linter making this change?

content/services/development.md
```diff
@@ -9,9 +9,180 @@ action:
...
+      width=744 ,
+      height=1611,
+      sizes='media'
+    )
+  )
+) %}
+
+{% call layout.grid('narrow-columns') %}
+
+{% call layout.block('item') %}
```
mirisuzanne, line 42: maybe call this `grid-item` or `column-item` or something that connects it to the `narrow-columns` grid? Alternately, consider using yaml data for these, like we do with icon blocks? We don't need to go that direction, but I tend to consider it when we otherwise need nested `call`s…

src/scss/config/_color.scss
```diff
@@ -82,6 +82,7 @@ $content-colors: (
   'action': var(--ccs-prime--fg3),
   'active': '#highlight',
   'icons': var(--ccs-special--fg1),
+  'markers': var(--ccs-special--fg1),
```
mirisuzanne, line 85: See comment below about a different way of handling this…

src/scss/patterns/_lists.scss
```diff
@@ -0,0 +1,58 @@
...
+    --list-padding--default: 1em;
+  }
+}
+
+// List Markers
+// ------------
+/// Lighter markers and smaller bullet points…
+/// @group type
+::marker {
+  color: var(--text-light);
```
mirisuzanne, line 33: I would use a stack here, with something like `var(--markers, var(--text-light))`, and then have `[data-block='fancy-list']` below set `--markers: var(--icons);`

I think that's more accurately the pattern here. In general we want markers based on text, but sometimes we want them to look more like our icon styles. By using a stack, we can also avoid the deep nesting below. (Could do something similar with font-size, by adding a custom property like `--marker-size`, and falling back to `1em` here.

src/scss/patterns/_sections.scss
```diff
@@ -69,7 +69,8 @@
 // ------
 
 [data-block] {
-  margin-block: var(--double-gutter);
+  margin-bottom: var(--double-gutter);
+  margin-top: var(--double-gutter);
```
mirisuzanne, line 73: `var(--block-spacer, var(--double-gutter))`? Then below we could set `--block-spacer: 0` on `[data-grid]`?

src/scss/patterns/_sections.scss
```diff
@@ -135,3 +147,28 @@
...
   --newline: inherit;
 }
+
+// Allows for smaller columns of text, like when used inside a media block
+[data-grid='narrow-columns'] {
+  --fit-min: calc(var(--grid-item) * 0.75);
+}
+
+// Container with centered top/bottom triangles
+.has-angles {
```
mirisuzanne, line 157: I think this could use a clearer name… but I can't think of one 😂

src/scss/patterns/_lists.scss
```diff
@@ -0,0 +1,54 @@
+@use '~accoutrement/sass/tools';
+@use '../config/scale';
+@use 'sass:map';
```
jgerigmeyer, line 3: Where is this being used?

content/services/development.md
stacyk: I had it in the yaml, but I think it is more readable when the content is the order it appears on the page in general (plus I tend do always mess something up when it comes to yaml). I know we have content in the front matter for many parts of the site. I am old school ¯\_(ツ)_/¯

mirisuzanne: lol, yeah - I agree it's a trade-off. I'm just careful about nested calls, because I never understand how they handle context. But this seems to work, so I guess you got it right! 😂

jgerigmeyer: (Sidenote: Nunjucks is buggy and I don't trust it 😬)

jgerigmeyer: @mirisuzanne A few of the bugs from v3.0.0 have been fixed by v3.2.0... but there are still a number of bugs with scoping, especially as it relates to `call` and `set` blocks:

- https://github.com/mozilla/nunjucks/issues/679
- https://github.com/mozilla/nunjucks/issues/912
- https://github.com/mozilla/nunjucks/issues/664
- https://github.com/mozilla/nunjucks/issues/1131

src/scss/patterns/_lists.scss
stacyk: it's not. I removed lines 2-3

content/services/development.md
stacyk: @jgerigmeyer are you asking to change this to using the yaml content instead of what is in here now? I'm not sure what you are saying.

mirisuzanne:
> @mirisuzanne can I not use /// @group type in different files? I see the title on the page is in the middle (split between two pages of css).

@stacyk you can - we just need to move that title to the "first place". It's not ideal, but ¯\\\_(ツ)_/¯

content/services/development.md
jgerigmeyer: @stacyk I wasn't saying that specifically, no. But I think we should all be vaguely aware of the existing bugs in Nunjucks, and hand-test any uses of `caller` (especially nested usage) or `set` to be sure it's actually working as we want.

stacyk:
OK, I fixed the typography styleguide page title order. Back to @mirisuzanne for final review.

PR #60, opened 2020-06-25
Improved Cascading-Colors UX

mirisuzanne:
## Cute animal pic

quizzical  bird

https://trello.com/c/twS6Ay2H/28-i-want-to-improve-the-view-color-options

## Description

## Steps to test/reproduce

Play with the color menu, and see how it works

## Show me

Screenshot 2020-06-30 17 50 07
Screenshot 2020-06-30 17 50 26
Screenshot 2020-06-30 17 50 42


mirisuzanne:
[I want to improve the view/color options](https://trello.com/c/twS6Ay2H/28-i-want-to-improve-the-view-color-options)

content/_includes/forms.macros.njk
```diff
@@ -99,6 +99,96 @@ params:
...
+    id=id,
+    required=required,
+    attrs={'data-radio': 'option'} | merge(attrs)
+  ) }}
+{% endmacro %}
+
+{# @docs
+label: fieldset
+category: Fields
+note: Create a
```
stacyk, line 156: Create a what? Don't leave me hanging 🧗‍♂️

src/scss/forms/_radios.scss
```diff
@@ -0,0 +1,49 @@
+@use '~accoutrement/sass/tools';
+@use '../config';
+
+/// # Radio Buttons
+/// @group input-radio
+
+/// @group input-radio
```
stacyk, line 7: duplicate lines

src/scss/patterns/_dropdown.scss
```diff
@@ -21,11 +21,11 @@
 [data-dropdown] {
   @include tools.z-index('dropdown');
   background: var(--bg);
-  border: thin solid var(--border-fade);
+  border: thin solid var(--active);
   box-shadow: 0 0 var(--quarter-shim) var(--shadow);
-  max-width: 14em;
+  max-width: calc(100vw - var(--page-margin) * 2);
```
stacyk, line 26: curious how we can up with this equation?

src/scss/forms/_radios.scss
mirisuzanne: These are actually two different blocks. The first is adding an h1 to the input-radio page, and the second is attached to the code block below. They could maybe use more detail, but they aren't duplicate.

src/scss/patterns/_dropdown.scss
mirisuzanne: the viewport-width, minus the two (left and right) page margins, should be the full content width.

mirisuzanne:
@stacyk I do know there are a lot of people who prefer low-contrast options for accessibility. It's best to provide a default that's high-contrast for the most common a11y need, but I think allowing that to be changed is also an a11y feature. I tried to limit the contrast bar to a reasonable range, while still allowing it to go low-contrast if desired.

src/scss/patterns/_dropdown.scss
stacyk: ah, so that is for small screens. Got it.

src/scss/forms/_radios.scss
stacyk: I sorta wondered if it was there for just the heading... ok, makes sense.

mirisuzanne:
[I want the light/dark menu to close on blur](https://trello.com/c/FGb5nJbu/88-i-want-the-light-dark-menu-to-close-on-blur)

mirisuzanne:
While addressing @stacyk's review, I also implemented a CSS-only click-anywhere-to-close trick that I saw github using on their menus. If that looks good, I think I've addressed all the comments here.

src/scss/config/_focus.scss
```diff
@@ -0,0 +1,12 @@
+// Focus
+// -----
+/// Consistent focus ring,
+/// is applied by default on element-focus,
+/// but some patterns (like radio-buttons) also require
+/// focus-ring applied to a sibling element.
+/// @group config-focus
```
jgerigmeyer, line 7: @mirisuzanne This is missing a group-mapping in `.sassdocrc`.

src/scss/patterns/_dropdown.scss
```diff
@@ -13,6 +13,26 @@
...
+
+    @include tools.before('') {
+      @include tools.position(0 fixed);
+      @include tools.z-index('bump');
+      cursor: default;
+      display: block;
+    }
+  }
+}
+
```
jgerigmeyer, line 35: 😍

content/_includes/icons/audio.svg
```diff
@@ -3,8 +3,6 @@
   xmlns="http://www.w3.org/2000/svg"
   aria-hidden="true"
   focusable="false"
-  width="20px"
-  height="20px"
```
jgerigmeyer, line 7: @mirisuzanne So. Herman runs these through SVGO, which by default removes the `viewBox` attr if it matches the height/width exactly. Then we run those optimized SVGs through gulp-svg-symbols, which breaks (with our custom sprite-sheet, which relies on `viewBox`) if `viewBox` is missing...

I'm not clear on the tradeoffs here, or when we do or don't need height/width. Probably this is a bug that should be fixed in Herman -- let's discuss.

mirisuzanne: well, they don't do exactly the same thing - but they do overlap a lot.

- `viewBox` is better for most things. We don't ever want to remove it. if we're choosing, it should absolutely be what we keep.
- I like _also_ having height/width, because they provide a default size for the icon. It's less essential, but it means they'll remain reasonably-sized until we explicitly resize them in CSS (which is simple when we want it).

You can see the difference here: https://cdpn.io/mirisuzanne/debug/002175ac77b34d172f1cdf769673c899

So… I'm ok with this, and it shouldn't hold up the  branch, but I don't love it as a restriction. Long term, I'm not sure we should have Herman do so much "optimization" as a documentation tool.

jgerigmeyer: Yeah, that makes sense to me. It'd be a quick fix to Herman to stop optimizing SVGs entirely...

src/scss/config/_focus.scss
jgerigmeyer: @mirisuzanne Given your comment on the SVGs, I think this is the only remaining thing before merge?

mirisuzanne:
@jgerigmeyer every little fix becomes a bigger fix. I think this also addresses https://trello.com/c/XTKW4RCc/87-heman-sizes-not-showing-preview-correctly-maybe-only-when-in-a-css-var now…

mirisuzanne:
[Heman sizes not showing preview correctly (maybe only when in a CSS Var?)](https://trello.com/c/XTKW4RCc/87-heman-sizes-not-showing-preview-correctly-maybe-only-when-in-a-css-var)

PR #67, opened 2020-07-06
Pagination

mirisuzanne:
## Cute animal pic

horse's mouth

https://trello.com/c/mGh7jzeU/83-i-want-pagination-for-resource-lists

## Description

## Steps to test/reproduce

- check out the `/blog/` (resources) page
- pagination links at top & bottom of list
- click through to sub-pages

## Show me

Screenshot_2020-07-06 Resources


content/blog/index.njk
```diff
@@ -1,16 +1,36 @@
...
   text: Browse by tag »
   url: /tags/
-summary: |
-  Learn how to write resilient CSS,
-  add WebSocket push notifications to a RESTful API,
-  and more with these articles, videos,
-  podcasts, and conference talks.
+pagination:
+  data: collections._sorted-posts
+  size: 15
```
mirisuzanne, line 9: Number of posts per page, which is kinda arbitrary…

mirisuzanne:
[I want pagination for resource lists](https://trello.com/c/mGh7jzeU/83-i-want-pagination-for-resource-lists)

test/js/pages.test.js
```diff
@@ -97,6 +98,15 @@ describe('page filters', () => {
...
   });
 
+  test('eventSort', () => {
+    const slugs = ['test1', 'events', 'test3'];
+    const years = ['2020', '2019', '2018'];
+    const actual = eventSort(collection4);
+
+    expect(actual.map((item) => item.fileSlug)).toEqual(slugs);
+    expect(actual.map((item) => item.year)).toEqual(years);
+  });
```
mirisuzanne, line 108: @jgerigmeyer @duggiemitchell I tried writing tests here, but they're failing, and I'm not sure why. Help? :)

content/_includes/post.macros.njk
```diff
@@ -609,3 +632,82 @@ params:
...
+      each instance should get a unique ID
+  level:
+    type: integer [1-6]
+    default: 2
+    note: Heading level
+#}
+{% macro page_nav(title, pagination, current=true, id='', level=2) %}
+{% if pagination %}
+{% set id = ['resource-nav-title', id] | join('-') %}
+<nav class="pagination" aria-labeledby="{{ id }}">
```
stacyk, line 665: I believe `aria-labeledby` is supposed to be `aria-labelledby` (2 `l`s).

content/_includes/post.macros.njk
```diff
@@ -609,3 +632,82 @@ params:
...
+  {% h level, {
+    'id': id,
+    'class': 'sr-only'
+  } %}
+    {{ title | mdInline | safe }}
+  {% endh %}
+  <p>
```
stacyk, line 672: I wonder if we might use the nav element here instead of a `<p>`? That's what I mostly see on some accessibility sites but there are always exceptions based on the other content on the page.

Also, I don't know if I am reading this correctly, https://www.w3.org/TR/wai-aria-practices-1.2/#general-principles-of-landmark-design but it seems like we could use the heading as the label since it is in the same container and it preceeds the nav... so, for instance 2019 Resources could be used instead of including a duplicate header for sr-only. Let me know if I am reading this wrong. 
image

mirisuzanne: - We are using the `nav`, just with a `p` inside it. We could remove the `p` (I think it's actually just in the way), but we don't need to replace it.
- oh, yeah, you're right about the header. :) I'll fix that.

mirisuzanne: @stacyk I think I've resolved all these issues, if you want to confirm

src/filters/pages.js
```diff
@@ -265,6 +265,20 @@ const pageYears = (collection, events = true) =>
...
+note: |
+  Sort pages based on either the page date,
+  or the most recently past event date.
+params:
+  collection:
+    type: array
+    note: containing 11ty page objects
+*/
+const eventSort = (collection) =>
+  pageYears(collection).sort((a, b) => a.sort > b.sort);
```
jgerigmeyer, line 280: It took me an hour to track down this bug... which is that `.sort()` expects a number, not a boolean. 😖 

I find the large number of small filters, many of which call each other, very difficult to follow and debug. I'd love to clean them up with an eye toward clarity of naming, combining where possible, and removing extra ones -- but obviously none of that should hold up merge/launch.

.eleventy.js
```diff
@@ -47,11 +47,14 @@ module.exports = (eleventyConfig) => {
...
   );
+  eleventyConfig.addCollection('_sorted-posts', (collection) =>
+    pages.eventSort(collection.getFilteredByTag('_post')),
+  );
   eleventyConfig.addCollection('oss', (collection) =>
     collection
       .getAll()
       .filter((item) => item.data.oss)
-      .sort((a, b) => b.date - a.date),
+      .sort((a, b) => b.sort - a.sort),
```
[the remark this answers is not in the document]
mirisuzanne: Oh, right - We'd have to run them through our `eventSort` or `pageYears` filter, which adds that property…

src/filters/pages.js
mirisuzanne: I'm very open to that. I'm not a JS expert, and I wrote a lot of these in a hurry. But usually when I split them it's because we ended up needing just one aspect of it somewhere. I'm curious to see how you'd approach it.

jgerigmeyer: Yeah, I didn't intend that as "this is bad" as it reads -- the JS is great, it's been built up somewhat rapidly, and could improve with a refactor. 👍

.eleventy.js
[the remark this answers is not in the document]
jgerigmeyer: Fixed -- though maybe we want a manual sort for OSS projects instead of most-recent-first? That would allow us to highlight the bigger or more active ones.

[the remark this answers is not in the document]
mirisuzanne: oh yeah, didn't see this was on OSS. I think we have a card for that already, just got de-prioritized: https://trello.com/c/dGgqSrQ9/58-i-want-featured-oss-projects

content/_includes/post.macros.njk
```diff
@@ -609,3 +632,82 @@ params:
...
+    {% if pagination.href.next %}
+      {{ utility.link_if('« older', pagination.href.next) }}
+    {% endif %}
+    {% if current %}
+      {{ page_current(pagination) }}
+    {% elif pagination.href.next and pagination.href.previous %}
+      |
+    {% endif %}
+    {% if pagination.href.previous %}
+      {{ utility.link_if('newer »', pagination.href.previous) }}
```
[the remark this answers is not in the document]
mirisuzanne: I played around with this a bit, and felt like it would take some more design considerations to make it look good. I think we should make a new card for it.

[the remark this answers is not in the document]
mirisuzanne: haha, ok ok. changing.

But I feel like if we go the other direction, we need to always show both.

mirisuzanne:
> If we're sorting the full blog list, I think we should probably sort the tag pages too (e.g. `/tags/article/`)?

@jgerigmeyer do you mean paginate the tag pages, or are you talking about sorting them by eventDate (in which case, I think only the tags with events would need it)?

The problem with paginating tag pages is… we use 11ty pagination feature to _create_ the tag pages, and we can't nest that. We'd have to do it manually. There's a simple approach to that, if we want to do it in certain instances, using our `index` feature to override the tag page. But if we want to automate it everywhere, that becomes more complex. I was hoping that could be a lower priority.

content/_includes/post.macros.njk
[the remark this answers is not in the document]
jgerigmeyer: Yeah, okay -- tho perhaps it's more straightforward now that we've switched the direction, so it could be on the far left? But I'm fine leaving that for a new card.

jgerigmeyer:
> > If we're sorting the full blog list, I think we should probably sort the tag pages too (e.g. `/tags/article/`)?
> 
> @jgerigmeyer do you mean paginate the tag pages, or are you talking about sorting them by eventDate (in which case, I think only the tags with events would need it)?

Yeah sorry, I meant "paginate" not "sort". 🤦 

> The problem with paginating tag pages is… we use 11ty pagination feature to _create_ the tag pages, and we can't nest that. We'd have to do it manually. There's a simple approach to that, if we want to do it in certain instances, using our `index` feature to override the tag page. But if we want to automate it everywhere, that becomes more complex. I was hoping that could be a lower priority.

Huh, okay -- I see that now. I'm fine making it a new story. Looks like it'd be highest priority for `/tags/article/` and `/tags/css/`.

PR #75, opened 2020-07-15
Ensure alt attrs on all images

mirisuzanne:
## Cute animal pic

## Description

Alt tags are currently missing from images that don't explicitly set them - but empty alt attributes should always be used instead of removing the attribute entirely. We can do this in `utility.show_attrs`/`utility.attr_if` by passing `true` when there is no `atr` value.

## Steps to test/reproduce

- See the hero image for `/talks/cascade-aligned/`


content/blog/2018/beyondtellerand.md
```diff
@@ -87,10 +87,6 @@ favorites:
   caption='Video from Beyond Tellerand'
 ) }}
 
-{{ contact.subscribe(
-  interests=['oddnews', 'beyondtellerand19']
-) }}
-
```
jgerigmeyer, line 93: @mirisuzanne Do we want to just remove this everywhere? Looks like it's only being used on one post (vueconf).

jgerigmeyer: @mirisuzanne LGTM pending your thoughts on ☝️

PR #87, opened 2020-09-01
[analytics] Add Google Analytics

stacyk:
Trello: https://trello.com/c/Cfim4Zem/104-add-tracking-ga-or-netlify


stacyk:
@jgerigmeyer @mirisuzanne are there any conditionals we want to include? I see there is an unfilitered code and filtered, but also a different property call OddBird dev. All three point to oddbird.net so I am not sure what to do with all of that info.

stacyk:
[Add tracking (GA or Netlify)](https://trello.com/c/Cfim4Zem/104-add-tracking-ga-or-netlify)

jgerigmeyer:
I also added Netlify analytics (https://app.netlify.com/sites/oddleventy/analytics) so we can compare.

PR #91, opened 2020-09-22
Add Django job posting

wlonk:
The jobs page index needs style TLC that I can't give it, but it points to the right things at least.

## Cute animal pic

pet serval standing up at the counter, checking out the coffee maker


jgerigmeyer (approved):
Interesting to compare this to the job description back in 2016 when we hired Kit (https://github.com/oddbird/oddsite/blob/e4465df3a3778b73b27ff7a88da8fdec9d24fcae/content/oddjob.rst). It's mostly the same, with a notable difference that we didn't mention Django at all in that one -- only "Python for the web, preferably in more than one framework (or no framework at all!)". We also did include paid OSS time in the other posting.

content/jobs/jobs.11tydata.yaml
```diff
@@ -0,0 +1,70 @@
+layout: base
+image:
+  svg: logos/oddbird-mark
+  type: angle
+  attrs:
+    style: '--align: center'
+
+# override hours as needed
+hours: 10 to 20
```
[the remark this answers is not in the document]
mirisuzanne: If you set that value on the posting, it should override the shared default. That's what the comment there is _meant to_ convey…

content/jobs/django2020.md
```diff
@@ -0,0 +1,31 @@
...
+sub: Leading our back-end development process
+date: 2020-09-23
+---
+
+OddBird is looking for a **Django Developer** to join the team,
+and we'd love to hear from you!
+
+## You:
+
+- Have experience working with Django, and Python more generally
```
jgerigmeyer, line 13: @wlonk @carljm What do you two think about whether we want to advertise for a Django dev vs a Python dev? I notice that in 2016 (for Kit's position) we didn't mention Django at all -- I'm assuming we do want Django experience though?

https://github.com/oddbird/oddsite/blob/e4465df3a3778b73b27ff7a88da8fdec9d24fcae/content/oddjob.rst

wlonk: Pro-Django:
 - All our existing projects are Django-based, and so it's pretty truthful about the work someone'd be doing at the outset.

Con-Django:
 - It limits the pool more; folks who might think "I know Python well, and I could learn Django, but I don't know it yet" might self-select out.

carljm: I might be inclined to mention Django but put more emphasis on Python (Eg "experience working with Python for the web, and ideally Django") But I don't feel strongly about it.

wlonk: Yes, that! That's the perfect phrasing, imo.

jgerigmeyer (approved):

SondraE (approved):
Looking great!
Should it be "...continental North _American_ timezone."?
I'm not sure.

mirisuzanne:
> Should it be "...continental North _American_ timezone."?
> I'm not sure.

I'm not either. Changed it to `based in a timezone from UTC-4 (Eastern) to UTC-8 (Alaska)`

PR #106, opened 2020-11-19
Add Ed to birds

jerivas:


jerivas:
[Move Kit to Emeritus & Add New Birds](https://trello.com/c/D7aSlxcr/111-move-kit-to-emeritus-add-new-birds)

PR #121, opened 2021-01-15
services page redesign

dvdherron:
## Cute animal pic
image

https://trello.com/c/dk4TGMcK/34-what-is-the-design-and-content-for-the-services-list-page

## Description
- [x ] Description is in Trello card

## Steps to test/reproduce
This is a layout change of the services page

## Show me
To come . . .

### DIFF
--- a/content/services/index.njk
+++ b/content/services/index.njk
@@ -13,11 +13,27 @@ summary: |
 ---
 
 {% import 'post.macros.njk' as post %}
+{% import 'embed.macros.njk' as embed %}
 
-{# placeholder for layout template changes #}
-
-{{ post.grid(
-  pages=services,
-  level=2,
-  collections=collections
-) }}
+{% for item in services %}
+  {% set page = collections.all | getPage(item) %}
+    <article {{ post.aria_label(page) }} data-work="summary">
+      <div class="offset-grid">
+        <div class="offset-item offset-media">
+            {% set image = page.data.image %}
+              {{ embed.img(
+                src=image.src,
+                alt=image.alt
+              ) }}
+        </div>
+        <div class="offset-item project-details">
+        <header class="post-header">
+          {% call post.banner(page, level=2,called=true) -%}
+           {{  page.data.title | mdInline | safe }}
+          {%- endcall %}
+        </header>
+          <p>{{ page.data.summary | md | elide | safe }}</p>
+        </div>
+      </div>
+    </article>
+{% endfor %}
--- a/content/services/index.njk
+++ b/content/services/index.njk
@@ -20,11 +20,11 @@ summary: |
     <article {{ post.aria_label(page) }} data-work="summary">
       <div class="offset-grid">
         <div class="offset-item offset-media">
-            {% set image = page.data.image %}
-              {{ embed.img(
-                src=image.src,
-                alt=image.alt
-              ) }}
+          {% set image = page.data.image %}
+            {{ embed.img(
+              src=image.src,
+              alt=image.alt
+            ) }}
         </div>
         <div class="offset-item project-details">
         <header class="post-header">
--- a/content/services/index.njk
+++ b/content/services/index.njk
@@ -23,7 +23,12 @@ summary: |
           {% set image = page.data.image %}
             {{ embed.img(
               src=image.src,
-              alt=image.alt
+              alt=image.alt,
+              attrs={
+                'style': {
+                'object-position': image.position
+              } | styles
+              }
             ) }}
         </div>
         <div class="offset-item project-details">
--- a/content/services/index.njk
+++ b/content/services/index.njk
@@ -14,6 +14,8 @@ summary: |
 
 {% import 'post.macros.njk' as post %}
 
+{# placeholder for layout template changes #}
+
 {{ post.grid(
   pages=services,
   level=2,
--- a/src/scss/patterns/_offset-grid.scss
+++ b/src/scss/patterns/_offset-grid.scss
@@ -43,3 +43,23 @@ $break: 'wide-break';
     padding-left: 0;
   }
 }
+
+//services page
+
+.offset-media {
+  display: block;
+  padding: 0;
+
+  &:nth-child(odd) {
+    padding: 0;
+  }
+
+  &:nth-child(even) {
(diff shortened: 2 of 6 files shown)


dvdherron:
[What is the design and content for the Services List page?](https://trello.com/c/dk4TGMcK/34-what-is-the-design-and-content-for-the-services-list-page)

dvdherron:
This is not really ready for a review, but I would love your thoughts on the direction here so far @mirisuzanne @SondraE. I tried to stick with some of the patterns on the work page. The images would obviously have to be adjusted in some way to fit into the shapes without appearing stretched or cropped in a weird way. I'm still figuring that part out.

mirisuzanne:
@dvdherron I think the direction is looking very cool, nice work! 

(The images aren't looking stretched for me, looks like we have `object-fit` in there already? I'm not seeing any functional issues at this point)

SondraE:
@dvdherron Looking really nice! 

The only thing I notice that could use work is image sizing (which I think is what you're referencing in your comment). 

- On my laptop screen size the Training and Workshops image stands out as being really tall even though the related text is shorter than the rest of the items. 
- When I look at the images at tablet and phone screen size, they look at bit broken. (See screenshot)

Screen Shot 2021-01-25 at 11 28 11 AM

content/services/index.njk
```diff
@@ -13,9 +13,27 @@ summary: |
...
+{% for item in services %}
+  {% set page = collections.all | getPage(item) %}
+    <article {{ post.aria_label(page) }} data-work="summary">
+      <div class="offset-grid">
+        <div class="offset-item offset-media">
+            {% set image = page.data.image %}
+              {{ embed.img(
+                src=image.src,
+                alt=image.alt
+              ) }}
```
mirisuzanne, line 27:
```suggestion
            {{ embed.img(
              src=image.src,
              alt=image.alt
            ) }}
```

indentation cleanup…

content/services/index.njk
```diff
@@ -13,9 +13,27 @@ summary: |
...
+        </div>
+        <div class="offset-item project-details">
+        <header class="post-header">
+          {% call post.banner(page, level=2,called=true) -%}
+           {{  page.data.title | mdInline | safe }}
+          {%- endcall %}
+        </header>
+          <p>{{ page.data.summary | md | elide | safe }}</p>
```
mirisuzanne, line 35: Could we get away with dropping `| elide` here? Are the summaries so long that we need to trim crop them?

src/scss/patterns/_offset-grid.scss
```diff
@@ -43,3 +43,24 @@ $break: 'wide-break';
...
+  display: block;
+  padding: var(--offset-media-padding);
+
+  &:nth-child(odd) {
+    padding: var(--offset-media-padding);
+  }
+
+  &:nth-child(even) {
+    padding: var(--offset-media-padding);
+  }
```
mirisuzanne, line 60: This feels like a lot of repetition to override the padding set in `layout.clip-base()` and `layout.clip()` mixins. Might be cleaner if we setup those mixins to allow a custom-property override? Or maybe we could override the existing `--clip-space` (I haven't looked closely at how that one works)?

src/scss/patterns/_offset-grid.scss
```diff
@@ -43,3 +43,24 @@ $break: 'wide-break';
...
+
+  img {
+    height: 100%;
```
mirisuzanne, line 63: Might get this by default by using `flex` or `grid` on the container, rather than setting the explicit height here? This works great, so not a real concern – just a slight preference for more implicit approaches where possible in CSS.

mirisuzanne (commented):
This is real nicely done - and required even less new code than I expected. Nice work!

content/services/index.njk
dvdherron: Yeah, "web design & development" and "research & concepting" are both almost twice as long as the other two services, so without the `elide` filter they look really tall. 

bild

bild

dvdherron:
> 
> 
> @dvdherron Looking really nice!
> 
> The only thing I notice that could use work is image sizing (which I think is what you're referencing in your comment).
> 
>     * On my laptop screen size the Training and Workshops image stands out as being really tall even though the related text is shorter than the rest of the items.
> 
>     * When I look at the images at tablet and phone screen size, they look at bit broken. (See screenshot)
> 
> 
> Screen Shot 2021-01-25 at 11 28 11 AM

Thanks! Yes, I do see how the images look word in that device size range. I'm playing with some fixes locally to get that right. I'll share soon.

src/scss/patterns/_offset-grid.scss
dvdherron: I made some tweaks to the custom properties in the layout mixins. These styles are related to the `offset-item` class too. Let me know if you think what I've done has simplified things. 

Still working on some of the rest of the review edits. Glad you like how it's going so far!

content/services/index.njk
mirisuzanne: makes sense - yeah, good call.

src/scss/patterns/_offset-grid.scss
```diff
@@ -49,14 +49,9 @@ $break: 'wide-break';
...
 
-  &:nth-child(odd) {
-    padding: var(--offset-media-padding);
-  }
-
-  &:nth-child(even) {
-    padding: var(--offset-media-padding);
+  & > * {
+    --offset-media-padding: 0;
   }
```
mirisuzanne: It looks to me like the outer setting (line 50) would cover it, since variables inherit - and this shouldn't be needed.

dvdherron: fixed

src/scss/patterns/_offset-grid.scss
dvdherron: setting the container to `flex` and setting `flex: 1 0 auto` on the items appears to work to get the images in the right place. 

I also added `object-position` to the images to get them centered in a better way. You don't fall off the screen anymore!

SondraE (approved):
looks really good to me!

mirisuzanne (approved):
looks great!

PR #137, opened 2021-04-05
[Inlinecode] Style inline code to stand out more

stacyk:
## Cute animal pic

glen-carrie-H7yvXA01MdY-unsplash

https://trello.com/c/DNEFIilE/120-as-a-design-snob-id-like-to-see-more-styles-to-distinguish-when-a-word-is-in-ticks

## Description
Inline code barely stood out. This helps fix that.

## Steps to test/reproduce
Check out the Adobe Case study or other Articles that contain code

## Show me
Screen Shot 2021-04-05 at 2 24 18 PM

Screen Shot 2021-04-05 at 9 21 16 AM


src/scss/patterns/_code.scss
```diff
@@ -29,6 +29,14 @@ pre {
   white-space: pre;
 }
 
+code:not([class^='language']) {
+  background: var(--ccs-special--bg5-fade);
+  border-radius: var(--quarter-shim);
+  color: var(--ccs-accent--fg1);
```
stacyk, line 35: @mirisuzanne I wasn't sure which colors are available to use for stuff this like so if this isn't allowed, let me know.

src/scss/config/_scale.scss
```diff
@@ -34,7 +34,7 @@ $font-sizes: (
   'h1': '#xlarge',
   'h2': '#large',
   'h3': '#medium',
-  'code': 0.9em,
+  'code': 0.8rem,
```
stacyk, line 37: @mirisuzanne this seems to even out the size of code since before hte paragraph code was pretty large and example code was smaller.

stacyk:
[As a design snob, I'd like to see more styles to distinguish when a word is in `ticks`](https://trello.com/c/DNEFIilE/120-as-a-design-snob-id-like-to-see-more-styles-to-distinguish-when-a-word-is-in-ticks)

src/scss/patterns/_code.scss
mirisuzanne: All the colors are available, but it might be good to assign them more semantic names before using them? For example, `var(--ccs-accent--fg1)` already has an alias as `var(--accent)` in the color config.

(and very strangely also `var(--feature-accent)`… maybe we have some cleanup to do 😅)

mirisuzanne (commented):
@stacyk left answers to your questions - otherwise looks good

src/scss/components/_resources.scss
```diff
@@ -66,7 +66,7 @@
   --type-flag: var(--bg);
   --type-flag-active: var(--callout);
   --shadow: transparent;
-  background-color: var(--feature-accent);
+  background-color: var(--feature-accent, var(--accent));
```
stacyk, line 69: I removed the color definition for --feature-accent in the config file, but added accent as the fallback. I probably don't need to keep --featured-accent here, but I thought it could come in handy in case we have certain features we may want to highlight, we can easily adjust that variable then.

stacyk (commented):
@mirisuzanne ready for a quick re-review

mirisuzanne (approved):
🎉

PR #139, opened 2021-04-14
[event-dates] Rethink how we show event post dates

stacyk:
pascal-muller-4EajIuUxgAQ-unsplash

## Trello
https://trello.com/c/C4AKsJpG/125-its-confusing-that-talks-are-listed-at-their-most-recent-date-but-show-the-created-date-on-the-resources-page

- [x] Remove "published date" on resources talk posts
- [x] Display 1-2 past talk dates as well (only upcoming are showing ATM)

## To test
https://deploy-preview-139--oddleventy.netlify.app/blog/

View upcoming and past "Talk" posts on /blog page to see the removed published date and a list of upcoming and/or past talk dates. 

image


content/_includes/post.macros.njk
```diff
@@ -236,7 +236,7 @@ params:
 {# @docs
 label: upcoming
 category: events
-note: Listof upcoming events from a post
+note: List of upcoming events from a post
```
stacyk, line 239: I'm not clear what works best here... I wondering it if makes sense to add a `past` event macro, or it makes more sense to use a conditional to check if it is future, and if not, show the past event dates with a heading or what not. Any thoughts as far as the approach @mirisuzanne @dvdherron?

content/_includes/post.macros.njk
```diff
@@ -355,8 +355,10 @@ params:
 
     {{ client(post.data.client) }}
 
-    {% if dateformat %}
-      {{ pubdate(post.date, post.data.end, dateformat) }}
+    {% if not post.data.events %}
+      {% if dateformat %}
+        {{ pubdate(post.date, post.data.end, dateformat) }}
+      {% endif %}
```
stacyk, line 361: I left a comment in Trello about whether we would want to also use this approach on the Event single view or if the published date is OK when not in a list.

stacyk:
[It's confusing that talks are listed at their most recent date, but show the created-date (on the resources page)](https://trello.com/c/C4AKsJpG/125-its-confusing-that-talks-are-listed-at-their-most-recent-date-but-show-the-created-date-on-the-resources-page)

content/_includes/post.macros.njk
```diff
@@ -245,15 +247,24 @@ params:
...
+  {%- if futureevents | length > 0 -%}
+    <h4>Upcoming</h4>
+  {%- elif pastevents | length > 0 -%}
+    <h4>Past</h4>
+  {%- endif -%}
+
+  {%- if events and (events | length > 0) -%}
     <ul class="next-event">
-      {%- for next in events -%}
+    {%- for next in events -%}
```
jgerigmeyer, line 266: When this is a list of past events, I think we want to reverse the order so that it shows the two _most recent_ dates instead of the two _oldest_ dates.

Screen Shot 2021-04-15 at 9 17 56 AM

content/_includes/post.macros.njk
```diff
@@ -268,6 +279,7 @@ params:
             ) }}
```
jgerigmeyer, line 279: When there's more than one date here ("past" or "upcoming"), it's really not clear what year each date is. In this example, the dates displayed are actually from 2015 and 2018:

Screen Shot 2021-04-15 at 9 17 56 AM

jgerigmeyer (commented):
I like this approach 👍

content/_includes/post.macros.njk
dvdherron: Commit [8a6e35c](https://github.com/oddbird/oddleventy/pull/139/commits/8a6e35c0873c5ab866b22bea59ca8bbbaecab0b7) adjusts the order of past events so that the most recent one appears first. @jgerigmeyer @stacyk 

It works best for talks/events where there were at least two in the same year. 

For past events where the two most recent ones took place in separate years, it makes it more clear that if we go this route we will have to adjust how we display years. 

Here, for the Agile Design Systems talk, the first event was in 2019 but the second one is from 2018
image

stacyk:
Added the year. If the talk spans multiple days, it shortens the month, otherwise it lists the month in full. 
image

One additional thing we may want to consider is an indication if there are more events that aren't listed, but I think that should be in a new story.

jgerigmeyer (approved):
Functionality looks good to me! I like @stacyk's idea of indicating if there are more dates other than the ones listed.

content/_includes/post.macros.njk
```diff
@@ -245,15 +247,45 @@ params:
...
+{% macro eventpreview(
   events,
   links=true
 ) %}
-  {%- set events = events | getFuture if events else none -%}
+  {%- set events = events -%}
   {%- set events = events | sort(attribute='date') -%}
-  {%- if events | length > 0 -%}
+  {%- set pastevents = events -%}
+  {%- set futureevents = events | getFuture  -%}
```
mirisuzanne, line 257: It seems like these two lists are mostly duplicated. We could maybe simplify by doing something like:

```njk
{%- set futureevents = events | getFuture -%}
{%- set isfuture = futureevents | length > 0 -%}
{%- set showevents = futureevents if isfuture else events | reverse -%}

  {%- if showevents | length > 0 -%}
    <h4>{{ 'Upcoming' if isfuture else 'Past' }}</h4>
    <ul class="next-event">
    {%- for next in showevents -%}
      etc...
```

That first sorts out what event list to show, and then only runs the output loop once.

content/_includes/post.macros.njk
mirisuzanne: I don't have strong feelings on this one way or the other. I think it's much less confusing on the detail page.

content/talks/wordpress-block-enhancements.md
```diff
@@ -3,23 +3,29 @@ title: WordPress Block Editor Enhancements
 sub: How to make the editing experience better
 author: stacy
 date: 2020-11-17
-venue: WordPress Naperville
-url: https://www.meetup.com/WordPress-Naperville/events/vzstwrybcpbwb/
+image:
+  src: talks/wp-blocks.jpg
+  alt: Stacy Kvernmo walks through the WordPress editor palette function code
```
mirisuzanne, line 8: I think we want height/width here, ideally?

mirisuzanne (approved):
This looks great! I left one suggestion for simplifying the logic, maybe?

stacyk:
@mirisuzanne I updated the image dimensions and post event macro. Thanks for the tip on that logic. We had that for loop once until we had to reverse the direction of the past events but this makes sense. :)

dvdherron (approved):
LGTM 🎉

jgerigmeyer (approved):
Looks good! @mirisuzanne You want to give final 👍 and merge?

mirisuzanne (approved):
🎉

PR #147, opened 2021-05-06
More consistent names for custom date formats.

jgerigmeyer:
@mirisuzanne This has no impact on the built site; just trying to use clearer and more consistent names for the formats (e.g. I don't know what "range" would mean, but "month_year" is clearer).

### DIFF
--- a/content/_includes/utility.macros.njk
+++ b/content/_includes/utility.macros.njk
@@ -208,7 +208,7 @@ params:
     }
   } -%}
 
-  {%- set ongoing = (format == 'month_year') and not end -%}
+  {%- set ongoing = (format == 'range') and not end -%}
   {%- set is_range = end and (start | string != end | string) and (not ongoing) -%}
 
   {%- if prefix %}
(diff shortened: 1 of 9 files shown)


content/_includes/utility.macros.njk
```diff
@@ -199,16 +199,16 @@ params:
...
     'same-m': {
-      'start': 'month_date',
-      'end': 'date' if (format == 'month_date') else 'date_year'
+      'start': 'month_day',
+      'end': 'day' if (format == 'month_day') else 'day_year'
     }
   } -%}
 
-  {%- set ongoing = (format == 'range') and not end -%}
+  {%- set ongoing = (format == 'month_year') and not end -%}
```
mirisuzanne, line 211: @jgerigmeyer changing range to month-year doesn't make a lot of sense to me. The purpose of that format is that it assumes a start and end date, and assumes the range is open-ended if there is no end date. "Range" seems much more accurate to that.

mirisuzanne (approved):
🎉

PR #158, opened 2021-06-14
Remove Bkgnd

SondraE:
## Cute animal pic
photo-sung-svi1Avjac3I-unsplash

[Trello Card](https://trello.com/c/t9T8JE7u/118-some-older-images-have-white-backgrounds-like-quarqnet-case-study-https-wwwoddbirdnet-work-quarqnet)

## Description
- Removed white background from CoachHub and QuartNet case study images
- Most of the images were in Dropbox, and I just resaved at pngs with transparent backgrounds
- Two images in the CoachHub case study were nowhere to be found, so I cropped the white background

## Steps to test/reproduce
- /work/coachhub
- /work/quarqnet/


dvdherron:
[Some older images have white backgrounds, like Quarqnet case study: https://www.oddbird.net/work/quarqnet/](https://trello.com/c/t9T8JE7u/118-some-older-images-have-white-backgrounds-like-quarqnet-case-study-https-wwwoddbirdnet-work-quarqnet)

dvdherron (approved):
This looks good to me. I wasn't sure though if I should be looking for more images with white backgrounds. I think my container query post has couple. Should those get changed as well?

jgerigmeyer (approved):

mirisuzanne:
@SondraE what do you mean by "Two images in the CoachHub case study were nowhere to be found, so I cropped the white background"?

mirisuzanne (commented):
The pages themselves look great - but some of these images are now _very_ large. Some of them over 1mb. I'm not sure that's reasonable. Maybe the new image tooling will help that, if we get it working? But I wonder if we can choose a background-color that we're ok with in both light & dark modes for those larger images?

stacyk:
> @SondraE what do you mean by "Two images in the CoachHub case study were nowhere to be found, so I cropped the white background"?

@mirisuzanne I am assuming the original layered Photoshop files perhaps?

stacyk:
implementation

@SondraE I just ran this one through https://tinypng.com/ and it went from 1.3MB to 294kb. It is a barely noticeable change in quality, but great savings

stacyk:
@dvdherron I think you could if you want. Yours could look like a frame was intentionally applied, although the second image isn't centered. I think you could get away with jpg still even if you wanted to crop it a bit so the file sizes would stay small. 
dark-theme-oddsite

dvdherron:
> @dvdherron I think you could if you want. Yours could look like a frame was intentionally applied, although the second image isn't centered.

@stacyk fixed the centering:
image


> I think you could get away with jpg still even if you wanted to crop it a bit so the file sizes would stay small.

and bumped down the file sizes on the pngs in the cq post. I think I remember when I originally posted these that making them .jpgs added a bunch of distortion once they were processed.

stacyk:
@dvdherron Your centered image looks good. I think we can use the https://oddbird.net/sample/macros/embed/#img-docs `img` macro instead of the figure if you ever have a need for a smaller image (as in, we won't force it to be 1280 I think) which also helps reduce distortion...  I am not saying that is what you want here, it may be an option for those images with text in them if we are watching file size and image quality. The old site used to have some classes that told the layout how big hte image should be (or rather, what gridline to extend it to) but I think we may have removed that.

dvdherron:
> 
> 
> @dvdherron Your centered image looks good. I think we can use the https://oddbird.net/sample/macros/embed/#img-docs `img` macro instead of the figure if you ever have a need for a smaller image (as in, we won't force it to be 1280 I think) which also helps reduce distortion... 

@stacyk I think the figure embed creates several file sizes too, right? 

480: 
image

960: 
image

and 1280 at the widest breakpoint:
image

stacyk:
> 
> @stacyk I think the figure embed creates several file sizes too, right?
> 


Yes... but it always shows them at the "extending past the column of text" size. We used to (probably still do if we don't use figure?) allow some images to be displayed/shown at a smaller size when it made sense like on Coachhub's case study https://www.oddbird.net/work/coachhub/  
Screen Shot 2021-06-18 at 11 36 30 AM

I think it would be helpful for us to document images a little more with examples of when to use which and show examples of that more than we currently do (I know we have some documentation, but I think we can add to it). I see the quote images in the CQ post are 660px wide and 757px wide (original file dimensions) and our figure macro, by using the srcset stuff, is stretching them to almost twice that size on wide screens making the text less clear. I know jpg's handle text worse than PNGs, which contributed to the distortion you mentioned before, but so does our marco forcing the image larger than it is. 

@mirisuzanne do you know if there is another setting or something that would not create the srcset image larger than the image size itself? I think that we wouldn't want the generated image to be larger than the original?

mirisuzanne:
@stacyk I think the srcset logic is already set to not increase an image size - but you can still stretch a small image with CSS, which is a different issue. I agree, we could use a lot more documentation around images & sizing/stretching & avoiding the strange extra paragraph tags.

SondraE:
> @SondraE what do you mean by "Two images in the CoachHub case study were nowhere to be found, so I cropped the white background"?

@mirisuzanne I couldn't find the original layered photoshop images or even the screenshots in dropbox, so I cropped out the white background instead of being able to remove that white background layer.

SondraE:
@stacyk I made all the pngs smaller via tinypng.com. Thanks!

mirisuzanne (approved):
@SondraE I think all the issues here have been resolved. Looks good!

PR #161, opened 2021-06-21
Default social image

SondraE:
photo-1547139586-e8a75f518421

## Steps to test/reproduce
/blog/oddsite.jpg
oddsite


SondraE:
@mirisuzanne I did do some simplifying, but left in more words than the previous default image. Does this work or should I simplify more, in your opinion?

mirisuzanne:
@SondraE It doesn't necessarily need to be simpler (you can see how [github packs in a lot of detail on a social image](https://twitter.com/TerribleMia/status/1404981473589948418)) - but I'm not immediately convinced that this is the most relevant information (or the best layout) for an isolated default image?

Is there a reason we want to use the header of the site, rather than a custom graphic with just the most essential information laid out custom?

content/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Software For Humans
+title: We Design & Develop Web Apps For Humans
```
mirisuzanne, line 2: Something like this makes sense to me, but remember the main place this shows up is in a potentially very small tab, so we want the most important/meaningful words at the front:

We des…

SondraE: In such a small context as you show in your screenshot, maybe the only change I would make would to reorder "Develop & Design." The other option I thought about would be "Web App Design & Development for Humans," but then only "Web App" would be likely to show up. I think it may be a toss up as to whether "Develop" or "Web App" is most important/meaningful. We could remove the word "We," but I think in a bigger context the phrase "Design & Develop..." sounds like we're telling someone else to do a thing instead of explain what our services are -- like maybe we help other people design and develop this own web apps for humans. Those are my musings. Thoughts? Do you like the new image?
oddsite

mirisuzanne: yeah, the image is great. Could even get that tagline back on it maybe - just so it's able to stand alone? Out of curiosity, I checked out the CSS Tricks default, and it's similarly simple title & tagline: 



I agree with your thoughts. "Web App Design & Development" doesn't seem like it would need the "we" - but I don't know if "web app" is a batter start than "we develop" or not? "Custom web applications"? …???

SondraE: @mirisuzanne I went with "Custom Web App Design & Development For Humans" as the title. Here's the updated picture.
oddsite

mirisuzanne (approved):
works for me.

I suppose we could consider moving "OddBird" to the front of the home page title if we wanted? That would have to happen in the template logic, but I can do it quick if you think it's a good idea.

SondraE:
@mirisuzanne Moving OddBird to the front of the homepage title sounds like a good idea to me.

mirisuzanne:
@SondraE how's that look?

jgerigmeyer (approved):

PR #164, opened 2021-06-24
Theme Color Metadata

SondraE:
guinea pig

## Steps to test/reproduce
- Safari Technology Preview browser
- Switching preferences between dark & light modes

## Screenshots
Dark Mode
Screen Shot 2021-06-24 at 3 00 40 PM

Light Mode
Screen Shot 2021-06-24 at 3 02 55 PM


SondraE:
[I want light/dark theme color metadata 🎨](https://trello.com/c/Qb84G7pc/141-i-want-light-dark-theme-color-metadata-%F0%9F%8E%A8)

mirisuzanne (approved):
That looks good to me. I'd be interested in feedback from more of the team (though it's not urgent since this is not yet supported in many stable browsers).

content/_layouts/base.njk
```diff
@@ -43,6 +43,9 @@
 
     {% include 'site/feed.njk' %}
 
+    <meta name="theme-color" media="(prefers-color-scheme: light)" content="#d9e4e7">
+    <meta name="theme-color" media="(prefers-color-scheme: dark)" content="#632645">
```
jgerigmeyer, line 47: I'm curious how these colors were chosen (esp for dark mode) -- that is, do they match other existing colors in our palette, or more "it looks good"?

jgerigmeyer (approved):
I mean, it's kinda cool 🤷

content/_layouts/base.njk
SondraE: Here you can see the two colors I chose. 
The dark raspberry color shows up on hover. 
The light blue color is the background on quotes.

Screen Shot 2021-06-25 at 12 06 45 PM

mirisuzanne: the light/dark blue may be more clear for our brand, since that's what we use for the logo?

stacyk (approved):
I was hoping that the colors in the editable palette would also control the theme color. Is that even possible? 

I have to admit I was not expecting the raspberry color. I think it stands out, but I also think the dark blue would be nice. I really like how Smashing Magazine's header sorta melts into the browser chrome. The raspberry color can get a little funky when paired with some greenish colors. In general, a blue hue tends to pair with a wider color palette. I don't feel strongly about it at the moment.

Screen Shot 2021-07-01 at 1 09 06 AM

I [added a new story](https://trello.com/c/uq3wMA1u/142-updated-favicons) to update/include some of the other favicon formats/files. I think we would want to consider how the icon would look on the backgrounds/themes that other authors choose on their sites. I'll add detail in said story.

SondraE:
@stacyk @jgerigmeyer You all were both a bit ambivalent about the raspberry color for dark mode. Here I've tried the dark blue color used in the bar at the top of the page. Since the top bar is semi-transparent, it doesn't blend seamlessly into the browser chrome on the left-hand side where the watercolor wash is. It's pretty subtle, though. Thoughts?
Screen Shot 2021-07-01 at 11 15 33 AM

stacyk:
It is subtle. I think that is OK @SondraE... It will never match thee bar on top since we allow visitors to edit it, unless that is, wee can change the value of the meta tag with a custom property.  Seems like we could, but isn't high priority yet. Maybe a fun retreat task.

report.20210701.111250.31070.0.001.json
```diff
@@ -0,0 +1,643 @@
+
```
jgerigmeyer, line 1: @SondraE Just a note to look at the files being committed before you commit/push to exclude these, and/or add `report.*.json` to your global `~/.gitignore` file.

dvdherron (approved):
The light mode and the the darker blue mode LGTM. The difference in color in color between the alert and the theme color is very subtle but I think it still looks OK.

PR #174, opened 2021-09-09
Add GH Actions for shadow repo

jerivas:
## Description
This configures GitHub Actions to build the site and push the `_site` folder to https://github.com/oddbird/oddleventy-built/. It will shadow the branch structure so diffing between `main` and any PR branch should show exactly what is changing in the final HTML.

_Keep in mind the `main` branch on `oddlevent-built` is currently empty so not really useful. Once this PR is merged `main` will receive all the contents of the `_site` folder. You can preview them in https://github.com/oddbird/oddleventy-built/tree/push-to-shadow_

## Steps to test/reproduce
No need to do anything locally. GH Actions will keep building and committing to the shadow repo on all pushes to all branches.

PR #175, opened 2021-09-10
Add diff link and branch deletion to shadow-repo actions

jerivas:
Follow up of #174


jerivas:
> View diff of compiled files: https://github.com/oddbird/oddleventy-built/compare/main..diff-link

@jgerigmeyer a small caveat with this is that the comment is generated immediately but the build can take several minutes to complete, which means the diff is empty while the project builds. It's safe to assume the diff is accurate once all checks have been completed though.

jgerigmeyer (approved):

PR #178, opened 2021-09-27
sort work sample list by end date

dvdherron:
## Description

- The `Work Samples` list should be sorted by end date (grouped by client). 

- Ongoing items should appear at the top of the work samples list. 

- This PR also changes the prefixes for ongoing client work items:
    For posts, items in a list of posts/tags, and the list of work samples, client dates take the 'since' prefix if they are ongoing.

## Steps to test/reproduce
- Navigate to the `/work/` page and note the sorting order of the first sample from each client. 
- The list should be in order of the end date in descending order, grouped by client. 
- If a work sample has no end date, it should come in descending order compared to the end or start date of its nearest sample. 

For items that end in the same year, use the work sample file to double check order. 

## Show me

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -79,12 +79,19 @@ module.exports = (eleventyConfig) => {
     collection
       .getAll()
       .filter((item) => item.data.client)
-
-      // @@@ to-do
-      // sort in descending order by end date
-      // if no end date, sort in descending order by date
-
-      .sort((a, b) => b.data.end - a.data.end),
+      .sort((a, b) => {
+        if (!a.data.end) {
+          if (a.date > b.data.end) {
+            return -1;
+          }
+        }
+        if (!b.data.end) {
+          if (a.data.end > b.date) {
+            return -1;
+          }
+        }
+        return b.data.end - a.data.end;
+      }),
   );
   eleventyConfig.addCollection('sample', (collection) =>
     collection.getAll().filter((item) => item.data.sample),
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -81,14 +81,10 @@ module.exports = (eleventyConfig) => {
       .filter((item) => item.data.client)
       .sort((a, b) => {
         if (!a.data.end) {
-          if (a.date > b.data.end) {
-            return -1;
-          }
+          return b.data.end ? -1 : b.date - a.date;
         }
         if (!b.data.end) {
-          if (a.data.end > b.date) {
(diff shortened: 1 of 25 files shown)


dvdherron:
[Sort work list by end-date.](https://trello.com/c/gOzn5yMJ/147-sort-work-list-by-end-date)

.eleventy.js
```diff
@@ -79,7 +79,19 @@ module.exports = (eleventyConfig) => {
...
+            return -1;
+          }
+        }
+        if (!b.data.end) {
+          if (a.data.end > b.date) {
+            return -1;
+          }
+        }
+        return b.data.end - a.data.end;
+      }),
```
jgerigmeyer, line 94: @dvdherron This is close, but I think probably we want ongoing clients (without an "end" date) to appear first in the list. That means we'll need to add end dates to old clients, but I can do that. I'll try something like:

```js
        if (!a.data.end) {
          return b.data.end ? -1 : b.date - a.date;
        }
        if (!b.data.end) {
          return 1;
        }
```

jgerigmeyer (commented):
I like this new sorting better. I think the only thing I notice that we should probably address is how ongoing projects are displayed in various places. Right now it's "on Jan 11" or just "2017", where I think "since..." would be better?

Screen Shot 2021-09-27 at 5 44 37 PM
Screen Shot 2021-09-27 at 5 47 19 PM
Screen Shot 2021-09-27 at 5 47 24 PM

dvdherron:
> 
> 
> I like this new sorting better. I think the only thing I notice that we should probably address is how ongoing projects are displayed in various places. Right now it's "on Jan 11" or just "2017", where I think "since..." would be better?
> 
> Screen Shot 2021-09-27 at 5 44 37 PM Screen Shot 2021-09-27 at 5 47 19 PM Screen Shot 2021-09-27 at 5 47 24 PM

Thanks for adding the end dates. Make this a lot simpler. 

Looking at the templates now to see what we could adjust got add "since" in the right places.

dvdherron:
Ok, @jgerigmeyer I think my changes cover all of these cases:

Post list / tags
image

Work sample list
image

Work/client post
image

jgerigmeyer:
> Ok, @jgerigmeyer I think my changes cover all of these cases:

This looks good!

jgerigmeyer (approved):

mirisuzanne (approved):
looks great!

PR #184, opened 2021-12-08
Upgrade JS deps.

jgerigmeyer:


package.json
```diff
@@ -45,67 +45,66 @@
...
-    "@rollup/plugin-node-resolve": "^13.0.5",
+    "@rollup/plugin-commonjs": "^21.0.1",
+    "@rollup/plugin-node-resolve": "^13.0.6",
     "@tommoor/remove-markdown": "^0.3.2",
     "@types/markdown-it": "^12.2.3",
-    "@typescript-eslint/parser": "^4.33.0",
-    "@yarnpkg/sdks": "^2.4.2-0",
+    "@typescript-eslint/parser": "^5.6.0",
+    "@yarnpkg/sdks": "^2.5.1-rc.5",
     "accoutrement": "^3.0.1",
```
jgerigmeyer, line 60: @mirisuzanne I think we could also upgrade Accoutrement (and Sass) in this PR?

mirisuzanne: yes please 👍🏼

jerivas (approved):
Not sure if I should be looking for anything in particular, but the diff here and on the built files look good to me 👍

package.json
jgerigmeyer: Doesn't need to hold up this PR.

PR #185, opened 2021-12-13
Remove JS job posting from display

patriciarealini:
polar bear cubs in the snow

## Steps to test/reproduce

## Show me


content/jobs/seniorjs2021.md
```diff
@@ -1,7 +1,7 @@
 ---
-position: open
+position: filled
 banner: Senior JavaScript Developer
-sub: Collaborating on our front-end development process
+sub: This position has been filled
```
mirisuzanne, line 4: I see that this is the language we've used on other postings - but since we're closing the applications before we've actually filled the position, it could be surprising to people still in the process? Maybe we should say something like "No longer accepting applications for this position"?

mirisuzanne (commented):
I think we can also remove `/jobs` from the nav?

content/jobs/seniorjs2021.md
patriciarealini: sounds good to me , i'll apply these changes

patriciarealini: resolved by https://github.com/oddbird/oddleventy/pull/185/commits/74405df3b2b881b5de05c2c7253e59852a23e8fa

mirisuzanne (approved):
looks great to me!

jgerigmeyer (approved):

PR #187, opened 2021-12-21
Update accoutrement

dvdherron:
## Steps to test/reproduce
- ` yarn ` to update dependencies
- `yarn serve` to run the site
- make sure Accoutrement is working in all of the places where we are using it. 
- Make sure the styleguide is showing correctly


dvdherron:
[Update Accoutrement](https://trello.com/c/7gHOVz27/151-update-accoutrement)

dvdherron:
@mirisuzanne @stacyk This still needs the updated Accoutrement map-compile fix (in-progress) for the style guide to display correctly. Otherwise, I think the site in general is ready for an initial review to see how it works using the updated Accoutrement.

src/scss/initial/_scale.scss
```diff
@@ -21,7 +22,7 @@ html {
   @include tools.tokens--(scale.$theme-sizes);
   @include tools.tokens--(scale.$page-sizes);
 
-  --extra: #{(tools.size('wide-page') - tools.size('page')) / 2};
+  --extra: #{math.div(tools.size('wide-page') - tools.size('page'), 2)};
   --main-min: calc(100vw - var(--page-margin) * 2);
```
mirisuzanne, line 26: Doesn't need to happen in this branch necessarily, but these seems like something that could move to the page-sizes map?

src/scss/layout/_hero.scss
```diff
@@ -7,11 +7,11 @@
   position: relative;
 
   &::before {
-    @include tools.position(0 -1em 0 -1em);
-
     box-shadow: 0 0 var(--quarter-shim) var(--shadow) inset;
     content: '';
+    inset: 0 -1em 0 -1em;
```
mirisuzanne, line 12: I think we can use the 2-value version for this?

```suggestion
    inset: 0 -1em;
```

mirisuzanne (commented):
Everything here looks great to me. Left a few possible cleanup notes, but only one of them is new to this branch.

src/scss/initial/_scale.scss
```diff
@@ -1,3 +1,4 @@
+@use 'sass:math';
 @use '../config/scale';
 @use '~accoutrement/sass/tools';
```
stacyk, line 3: @mirisuzanne @dvdherron I am curious to learn more about how we want to structure this moving forward. I know in TimeDesigner we used the `config` folder when we needed it (or tools) and forwarded & used `tools` in the config index. Are we not doing that here because this site was already set up with Sass modules and we are doing a quick update to Accoutrement? Or is there more that we consider when deciding how to structure that?

stacyk (approved):
@dvdherron looks good to me. I looked at the compiled site and through the code. 👍

src/scss/initial/_scale.scss
mirisuzanne: I don't think we've discussed it in detail, so I think it's the 'quick update' that's happening here. I'd be happy to talk through what pattern we want to use, maybe set up a meeting to discuss?

src/scss/layout/_hero.scss
dvdherron: @mirisuzanne changed this to the 2-value version

src/scss/initial/_scale.scss
dvdherron: @mirisuzanne I made a new card to take care of this. if `map-compile` gets merged into this branch soon I can look into relocating those variables afer that's done.

src/scss/initial/_scale.scss
dvdherron: @stacyk @mirisuzanne I put a date on the calendar this Thursday. Maybe then we can touch on this and some of the remaining Accoutrement update/testing work?

dvdherron:
@mirisuzanne @stacyk I've updated this so that it's using the `compile-*` functions in the right places. I think I've covered all of the places where it was needed (scale, color, fonts) and Herman seems to be previewing everything correctly. 

If we want to re-do the configuration now, I can make those changes here too. Or I can I can make a new card and get to that after updating the rest of the Accoutrement docs is done.

mirisuzanne:
> If we want to re-do the configuration now, I can make those changes here too. Or I can I can make a new card and get to that after updating the rest of the Accoutrement docs is done.

@dvdherron I think a new card would be good - no reason to hold up this branch

mirisuzanne (approved):
lgtm!

src/scss/config/_scale.scss
```diff
@@ -87,12 +81,7 @@ $theme-sizes: (
 );
 
 @include tools.add-sizes($theme-sizes);
-@include herman.add(
-  'sizes',
-  'theme-sizes',
-  $theme-sizes,
-  get-function('size', $module: 'tools')
-);
+@include herman.add('sizes', 'theme-sizes', tools.compile-sizes($theme-sizes));
```
jgerigmeyer, line 84: Something seems to be causing an unwanted change to how the `border-width` size is displayed:

Before (https://www.oddbird.net/styleguide/config-scale#variable--theme-sizes):

Screen Shot 2022-01-24 at 4 40 16 PM

After (https://deploy-preview-187--oddleventy.netlify.app/styleguide/config-scale#variable--theme-sizes):

Screen Shot 2022-01-24 at 4 40 12 PM

jgerigmeyer (changes requested):

src/scss/config/_scale.scss
mirisuzanne: Interesting. Both are wrong…

Herman doesn't currently support visualizing named lengths like this. I filed a card to fix it in Herman: https://trello.com/c/gAFt8dMk/59-herman-how-should-length-keywords-like-a-medium-border-width-be-previewed

stacyk (approved):
This all looks good to me. Yeah, that border width (keyword) is not new so I think this is good to go.

src/scss/config/_scale.scss
dvdherron: @jgerigmeyer Hmm, before this I think it was using the font-size reference value for `medium` instead of the CSS keyword. 

@mirisuzanne until this is supported in Herman, should we replace the keyword value with just a raw length value (3px)?

jgerigmeyer (approved):

src/scss/config/_scale.scss
mirisuzanne: I don't feel like we need to adjust our styles to match what Herman supports, but I also don't think it hurts much if you want. ¯\\\_(ツ)_/¯

PR #191, opened 2022-01-28
experiment with netlify caching

jgerigmeyer:


jgerigmeyer:
This doesn't actually change our caching strategy, but should slightly improve build times by telling Netlify not to first install another version of Yarn.

jerivas (approved):
🚀

PR #200, opened 2022-03-02
Upgrade deps

jgerigmeyer:


jgerigmeyer:
@mirisuzanne Not much changes here in practice. Images are run through a newer optimizer, and similarly the syntax highlighting engine is upgraded. Otherwise it's mostly doc updates for True and Accoutrement.

mirisuzanne (approved):
lgtm

PR #204, opened 2022-04-13
Adding search to top bar

oluoluoxenfree:
## Description

Algolia search for the eleventy site. Isn't currently opening a new page, need to see if that's possible with instant search. Also needs to highlight the search term. Also need to truncate the content of some posts, but unsure best way to do that. 

## Show me
search-narrow-dark
search-narrow-dark-expanded
search-narrow-light
search-wide-light-expanded
search-wide-light

search-wide-dark

## Additional things to note:
- The buttons site-wide are all squared off as the rounded corners didn't really fit with the rest of the styles. It was really apparent with the search input right next to the light/dark button. The only buttons that I believe aren't yet converted is the light/dark button group buttons, but we have a card to address this.
- The input styles were changed site-wide to include the accent color outline as shown in the original search mockup. https://xd.adobe.com/view/9d5eccab-49f0-4af3-7f1e-1588453f8878-eb6f/
- The narrow search functionality of the search doesn't really work as I expect it to. I hid the clear input button as that didn't only clear the input, but it also closed the form altogether. The wider screen version will clear your text,  but the focus is still in the input which is what I would expect.
- The submit button also closes the form for some reason on the small screen version.
- I didn't match the mockup and tried to find a balance between Algolia's styles with some necessary overrides to match our site. There are still a lot of strange looking overrides since Algolia has a lot of opinionated css.


oluoluoxenfree:
[How do  we want searching to look and work?](https://trello.com/c/GozqBHVi/20-how-do-we-want-searching-to-look-and-work)

jgerigmeyer:
@stacyk @SondraE This is now setup with the easiest option, and ready for review. If this is good enough, we can add our own custom styles and call it done. If this doesn't provide us with the functionality we want, then we can move to the next option, which is a similar widget but where we can define our own templates/options (and add filters, etc).

https://deploy-preview-204--oddleventy.netlify.app/

With the current approach, we can pretty much only customize colors (https://www.algolia.com/doc/tools/crawler/netlify-plugin/front-end/#theme) plus whatever custom styles we want to add as overrides -- or I guess we can remove their styles entirely if @stacyk would prefer that.

@stacyk While working on this, you can add `debug: true` to the options passed into `algoliasearchNetlify` in `src/js/index.js` -- that will allow the dropdown to stay open instead of auto-closing on blur, so you can inspect the markup etc. Let me know if you have any other questions, or if it would be easier to look at this together on a call.

SondraE:
@stacyk @jgerigmeyer The way this works generally looks good to me. I found one thing that might be confusing if I weren't so familiar with oddsite. I typed in Second Measure. When I clicked on the top search result I went to the Second Measure detail page which was perfect, but when I clicked on the second search result I was taken to the top of the Work list page. In order to actually find the term I searched for on the Work page, I needed to search again using Command/F. I can understand if there's nothing we can do to improve this, though.

mirisuzanne:
@SondraE I think only Chrome allows you to link to a random bit of text somewhere in the document. So there's likely nothing we can do about it, at least in most browsers.

stacyk:
What if it is a new anchor link? Can the search plugin scan for those links maybe?

jgerigmeyer:
> What if it is a new anchor link? Can the search plugin scan for those links maybe?
> 
> I should clarify though, I think this flow is pretty typical. I see the title of the page in the results and am going to assume the search term is somewhere on the page, but I may have to scroll/look a bit to find it.

@SondraE @stacyk @mirisuzanne I updated our Crawler settings to use a "hierarchical" schema (https://www.algolia.com/doc/tools/crawler/netlify-plugin/extraction-strategy/#hierarchical-schema), so it now links to headings within pages. It does mean more results overall, but maybe that's okay? Play with it and see what you think.

SondraE:
@jgerigmeyer @stacyk 
I really do like how the new hierarchical schema takes me to Second Measure on the Work page! 

I noticed that when I search with the word development, there is no search result for the Service page header "web design & development." That result is a lot more relevant than the "tag: business development" result that does make the list. I'm not sure if this is a result of the new schema or not because I didn't search development before. (Are we limiting the number of results that are displayed in the drop down?)
I find that I don't frequently use marketing websites' search because the results tend not to be very relevant in just this way. I wonder if this is just the nature of search plugins. Is there anything we can do to improve the relevance of the results, maybe even for just a few of the terms we think are the most important?

As Stacy pointed out, people may not find the original flow confusing because it is pretty typical. We can move back to that, if needed, especially if it improves the relevance of the results.

jgerigmeyer:
> Is there anything we can do to improve the relevance of the results, maybe even for just a few of the terms we think are the most important?

Yes, there are obviously default algorithms for this, but we have lots of control over customizing the relevance/rankings of the results (I think?). I added everyone to the Algolia project, where you can check out an overwhelming number of settings. I'm happy for someone to dig into all the various intricacies of search results.

- https://www.algolia.com/apps/WVUJ3O7KS6?section=search
- https://www.algolia.com/doc/guides/managing-results/relevance-overview/
- https://www.algolia.com/doc/guides/managing-results/must-do/custom-ranking/

> Are we limiting the number of results that are displayed in the drop down?

Yes, it is limited to 5 results by default. I changed that to 10, but let me know what you prefer.

stacyk:
> > Are we limiting the number of results that are displayed in the drop down?
> 
> Yes, it is limited to 5 results by default. I changed that to 10, but let me know what you prefer.
> 

@jgerigmeyer @SondraE  Maybe it depends on our ability to customize the results. I will take a look at the plugin documentation and login with the invitation you sent Friday. As of now, searching the word "design" shows me the top 4 hits are on the homepage, and one includes the alt text (or filename/url) of the timedesiger logo (which may be confusing since "Tegy" is what is shown). If we can clean up these results then 5 might not be too few. I'd also like to see if there's maybe a "load more" type of option in case.

I do like having the ability to link to the anchor links but between the two search bars at the moment, the left most "default" one gave me better results (mostly because of the duplicate links to a single page).

package.json
```diff
@@ -42,6 +42,9 @@
     "prod": "run-s update-oss-docs build",
     "checklinks": "yarn dlx hyperlink --canonicalroot https://www.oddbird.net/ -r -p --skip \"_site/feed.atom\" --skip \"docs.djangoproject.com\" --skip \"/docs/\" --skip \"codepen.io\" --skip \"github.com\" --skip \"caniuse.com\" --skip \"youtu.be\" --skip \"stackoverflow.com/users/\" _site/index.html"
   },
+  "dependencies": {
+    "@algolia/algoliasearch-netlify-frontend": "^1.0.14"
+  },
```
stacyk, line 47: @jgerigmeyer I had a merge conflict here so I accepted the latest on main and then added in the dependencies with only the search package... wasn't sure if you wanted to move accoutrement, ccs and css remedy here like it was in this branch before this commit or move search to the devDependencies list?

SondraE (approved):
Looks and works as I expected from Stacy's walkthrough. Love it!

jgerigmeyer (approved):
🚀

src/scss/config/color/_content.scss
```diff
@@ -9,3 +9,10 @@ $shadow: hsla(var(--ccs-h--neutral), var(--ccs-s--fg6), var(--ccs-l--fg6), 25%);
 $action: var(--ccs-prime--fg3);
 $active: $highlight;
 $icon: var(--ccs-special--fg1);
+$scroll-panel-bg: hsl(
+  var(--ccs-h--prime),
+  var(--ccs-s--bg4),
+  var(--ccs-l--bg5)
+);
+$scrollbar-track: var(--ccs-prime--bg4-fade);
+$scrollbar-thumb: var(--ccs-prime--bg6-fade);
```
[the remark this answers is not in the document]
stacyk: no, I originally had it in the bg colors map, but when I brought the sass coffin stuff into this branch, I think I changed my mind re: what a bg color is and forgot to update it when I moved it to content. I'll fix.

content/_includes/site/search.njk
```diff
@@ -1,33 +1 @@
...
-    </button>
-    <ul
-      data-navlist="search"
-      data-dropdown="links"
-      style="--right: 0;"
-      aria-expanded="false">
-    </ul>
-  </div>
-</form>
+<div id="search" data-form="search"></div>
```
mirisuzanne, line 1: Might not be worth an entire include, but 🤷🏻‍♀️

src/scss/components/_search.scss
```diff
@@ -0,0 +1,250 @@
+@use 'sass:meta';
+@use '../config';
+
+@layer search {
```
mirisuzanne, line 4: 👀🥳 layers!

src/scss/config/color/_content.scss
mirisuzanne, line 18: Wow it's hard to inspect what these actually do. As soon as you open the inspector, all the search results disappear? I don't know if that's good or bad in terms of the actual UX. But I'm really curious: is Algolia actually hijacking scroll somehow, or are they using some experimental css to style the default scrollbars, or something else?

src/scss/layout/_menu.scss
```diff
@@ -1,7 +1,7 @@
 [data-layout='menu'] {
-  align-items: center;
   display: flex;
   font-size: var(--small);
+  gap: var(--shim) var(--gutter-plus);
```
mirisuzanne, line 4: Seems like we also need flex-wrap set here, in order for this to work as intended? Right now they stay side by side, and cause overflow on small screens.
Screen Shot 2022-08-30 at 2 32 40 PM

mirisuzanne (commented):
Looks great! Left a few small questions.

src/scss/config/color/_content.scss
stacyk: @mirisuzanne you can test it easier with this pro tip:

> @stacyk While working on this, you can add debug: true to the options passed into algoliasearchNetlify in src/js/index.js -- that will allow the dropdown to stay open instead of auto-closing on blur, so you can inspect the markup etc. Let me know if you have any other questions, or if it would be easier to look at this together on a call.

https://github.com/oddbird/oddleventy/pull/204#issuecomment-1100216083

mirisuzanne: thanks, looks good!

PR #205, opened 2022-04-19
change Sass configuaration

dvdherron:
## Steps to test/reproduce
- Sass files should be loading Accoutrement from a @forwarded configured version (from  `tools` or `config`
- Check the Styleguide to make sure previews and content are loading as they should
- Check the site diffs to see if any layout issues have been introducted

## Show me

### DIFF
--- a/src/scss/config/color/_index.scss
+++ b/src/scss/config/color/_index.scss
@@ -1,0 +1,2 @@
+@forward 'brand';
+@forward 'content';
--- a/src/scss/initial/_colors.scss
+++ b/src/scss/initial/_colors.scss
@@ -1,4 +1,6 @@
+@use 'sass:meta';
 @use '../config';
+@use '../config/color/content';
 @use '~cascading-color-systems' as ccs with (
   $hues: config.ccs('hues'),
   $saturation: config.ccs('saturation'),
@@ -24,7 +26,7 @@
 [data-ccs-colors] {
   @include config.tokens--(config.$ccs-colors);
   @include config.tokens--(config.$background-colors);
-  @include config.tokens--(config.$content-colors);
+  @include config.tokens--(meta.module-variables('content'));
 
   --img-bright--base: #{ccs.if-mode($light: 50%, $dark: 40%)};
   --img-bright--default: calc(var(--img-bright--base) + var(--ccs-l));
--- a/src/scss/initial/_colors.scss
+++ b/src/scss/initial/_colors.scss
@@ -1,5 +1,4 @@
-@use '../config/color' as config;
-@use '~accoutrement/sass/tools';
+@use '../config';
 @use '~cascading-color-systems' as ccs with (
   $hues: config.ccs('hues'),
   $saturation: config.ccs('saturation'),
@@ -23,9 +22,9 @@
 
 /// @group init-colors
 [data-ccs-colors] {
-  @include tools.tokens--(config.$ccs-colors);
-  @include tools.tokens--(config.$background-colors);
-  @include tools.tokens--(config.$content-colors);
+  @include config.tokens--(config.$ccs-colors);
+  @include config.tokens--(config.$background-colors);
+  @include config.tokens--(config.$content-colors);
 
   --img-bright--base: #{ccs.if-mode($light: 50%, $dark: 40%)};
   --img-bright--default: calc(var(--img-bright--base) + var(--ccs-l));
--- a/src/scss/components/_events.scss
+++ b/src/scss/components/_events.scss
@@ -1,11 +1,11 @@
-@use '~accoutrement/sass/tools';
+@use '../config';
 
 .event {
   margin-bottom: var(--newline);
 }
 
 .event-meta {
-  @include tools.font-family('alt');
+  @include config.font-family('alt');
 
   font-size: var(--small);
 }
@@ -15,7 +15,7 @@
 }
 
 .event-links {
-  @include tools.wrap-content('[' ']');
+  @include config.wrap-content('[' ']');
 
   font-size: var(--small);
   font-style: italic;
--- a/src/scss/components/_logo-grid.scss
+++ b/src/scss/components/_logo-grid.scss
@@ -1,5 +1,3 @@
-@use '~accoutrement/sass/tools';
-
 .logo-grid {
   --column: content;
   --fill: var(--feature);
--- a/src/scss/components/_post.scss
+++ b/src/scss/components/_post.scss
@@ -1,4 +1,4 @@
-@use '~accoutrement/sass/tools';
+@use '../config';
 
 [data-post] {
   --post-margin: var(--gutter);
@@ -32,7 +32,7 @@
 }
 
 [data-post='link'] {
-  @include tools.transition('pop');
+  @include config.transition('pop');
 
   --hero-height: var(--grid-hero);
   --item-title: 1rem;
@@ -58,7 +58,7 @@
     display: block;
 
     &::after {
-      @include tools.z-index('bump');
+      @include config.z-index('bump');
 
       content: '';
       display: block;
@@ -77,7 +77,7 @@
 // --------
 
 .type-flag {
-  @include tools.font-family('alt');
+  @include config.font-family('alt');
 
   background: var(--highlight);
   color: var(--bg);
--- a/src/scss/components/_resources.scss
+++ b/src/scss/components/_resources.scss
@@ -1,4 +1,4 @@
-@use '~accoutrement/sass/tools';
+@use '../config';
 
 // pagination
 
@@ -17,7 +17,7 @@
     [title-end];
   margin: var(--double-gutter) 0;
 
-  @include tools.above('wide-break') {
+  @include config.above('wide-break') {
     --justify-nav: initial;
 
     grid-template-columns:
--- a/src/scss/components/_work-grid.scss
+++ b/src/scss/components/_work-grid.scss
@@ -1,4 +1,4 @@
-@use '~accoutrement/sass/tools';
+@use '../config';
 
 [data-work] {
   --column: full;
@@ -6,7 +6,7 @@
   [data-work='summary'] + & {
     --pull-sample: var(--spacer);
 
-    @include tools.above('wide-break') {
+    @include config.above('wide-break') {
       --pull-sample: calc(var(--spacer) * -1);
     }
   }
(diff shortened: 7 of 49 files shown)


dvdherron:
[I want to reconfigure the Sass configuration pattern](https://trello.com/c/f4xbWI4m/158-i-want-to-reconfigure-the-sass-configuration-pattern)

dvdherron:
@mirisuzanne @stacyk This is mostly changing the way we load Accoutrement in our files, similar to what we in [Time Designer](https://github.com/oddbird/time-designer/pull/1047). 

In the Styleguide, I'm wondering how you two feel about the ordering. Is there a way to specify when each item appears if we want the order to be Color Config --> Brand Colors -->> Background Colors . . .etc in a similar order to the current styleguide?

I haven't split `animation` or `scale` yet but if that's something we want to do I can start on that here or split those off into separate stories.

src/scss/initial/_colors.scss
```diff
@@ -23,9 +26,9 @@
 
 /// @group init-colors
 [data-ccs-colors] {
-  @include tools.tokens--(config.$ccs-colors);
-  @include tools.tokens--(config.$background-colors);
-  @include tools.tokens--(config.$content-colors);
+  @include config.tokens--(ccs-defaults.$ccs-colors);
```
mirisuzanne, line 29: If we break the color partials into individual variables, then we can use `meta.module-variables` here to pass along a map of those variables. Otherwise, we don't seem to be actually making use of `sass:meta` in this file yet.

stacyk (commented):
@mirisuzanne I made some changes to address your review and make this project more inline with how we have OddBooks structured. I believe we have to use CSS variables for some of the CCS stuff to work, and Herman doesn't have access to the definitions of those colors right? I added and removed and re-added the 

```
@include herman.add(
  'colors',
  'x-colors',
  tools.compile-colors(meta.module-variables('all-other-color-files'))
);
```

for all the content, background, and ccs default color files (brand uses color values to define the colors instead of custom properties so those are fine). I know the maps would display the color name and value before this change even though it didn't show a preview. If I add that block above for each color file, the colors will show up like a preview, list the color name and the variable value... but all of them are empty. Is there a way load the styleguide.css file like it does when you use an example like the focus-ring styleguide page? https://www.oddbird.net/styleguide/config-focus

src/scss/config/color/_ccs-defaults.scss
```diff
@@ -0,0 +1,11 @@
+/// ## CCS Defaults
+/// ------------
+/// The Cascading Color System has a few settings
+/// that we can most easily override with CSS variables
+/// @colors ccs-colors
+/// @group config-color
+
+/* stylelint-disable-next-line dollar-variable-pattern */
+$ccs-neutral--theme: var(--ccs-h--prime);
```
stacyk, line 9: @jgerigmeyer @mirisuzanne the linter yells at me even after I added this line so I am not sure what to do instead?

jgerigmeyer: Fixed in https://github.com/oddbird/oddleventy/commit/fc17e225d2eb008cc0b5208e327db137ea0de80f

src/scss/config/color/_ccs-defaults.scss
```diff
@@ -3,10 +3,9 @@
...
-/// @type map
 /// @group config-color
-$ccs-colors: (
-  'ccs-neutral--theme': var(--ccs-h--prime),
-  'ccs-background': var(--bg),
-  'ccs-color': var(--text),
-);
+
+/* stylelint-disable-next-line scss/dollar-variable-pattern */
+$ccs-neutral--theme: var(--ccs-h--prime);
```
mirisuzanne, line 9: We might want to update our `scss/dollar-variable-pattern` setting to allow this pattern, or else do a find/replace to remove the double-dashes?

src/scss/initial/_colors.scss
```diff
@@ -19,13 +19,9 @@
...
 /// @link https://www.oddbird.net/cascading-colors/
 ///   Cascading Color Systems
-/// @group init-colors
-
 /// @group init-colors
 [data-ccs-colors] {
-  @include tools.tokens--(config.$ccs-colors);
-  @include tools.tokens--(config.$background-colors);
-  @include tools.tokens--(config.$content-colors);
+  @include config.colors--;
```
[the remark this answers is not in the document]
mirisuzanne: I don't know of any reason we would need to, as long as it's working. :)

mirisuzanne (approved):
looks good! (noted an update we could make to the linter (or the… lintee?), but doesn't seem urgent)

src/scss/config/color/_ccs-defaults.scss
stacyk: @jgerigmeyer do you have a preference? I would need a few days to decipher this line (@jerivas here's one example of when that regex lesson would come in handy): 
`^([_|-]*[a-z][a-z0-9]*)(-[a-z0-9]+)*-*$`

mirisuzanne: Playing around in RegExr a bit, we could allow:

- 1 or 2 dashes in the middle of a name: `^([_|-]*[a-z][a-z0-9]*)(-{1,2}[a-z0-9]+)*-*$`
- any number of dashes in the middle: `^([_|-]*[a-z][a-z0-9]*)(-*[a-z0-9]+)*-*$`

Currently we're allowing any number at the start and end of the name.

jerivas: Oh nice! TIL about RegExr. I was previously using https://regex101.com/r/vx05jF/1 but the explainer on RegExr is much better ✨

jerivas: You can save any regex and get a link BTW regexr.com/6ri58

src/scss/config/color/_index.scss
```diff
@@ -0,0 +1,32 @@
...
+  'colors',
+  'ccs-colors',
+  tools.compile-colors(meta.module-variables('ccs-defaults'))
+);
+
+@include herman.add(
+  'colors',
+  'content-colors',
+  tools.compile-colors(meta.module-variables('content'))
+);
```
jgerigmeyer, line 32: Most of these aren't showing up in the Herman docs correctly. It looks like before we were only adding `brand-colors` -- should we stick with that? We were also documenting the actual maps before; not sure if that change was intentional.

Screen Shot 2022-08-09 at 1 49 30 PM

src/scss/initial/_colors.scss
```diff
@@ -19,13 +19,9 @@
 ///   Accoutrement Variables
 /// @link https://www.oddbird.net/cascading-colors/
 ///   Cascading Color Systems
-/// @group init-colors
-
```
jgerigmeyer, line 23: Removing this switches the order of the heading and the map. Should we add this back in?

Screen Shot 2022-08-09 at 3 30 10 PM

src/scss/config/color/_index.scss
stacyk: @jgerigmeyer Since Herman doesn't have access to the variables these will all be blank. It is nice to see the value (or variable name) for each color even if we don't see the preview ... whether it is in this form or in the sass map form, but we aren't defining colors in a map anymore, not sure if it makes sense to show the map?

Preferably, we'd see the color in the preview box :)  

Is this possible what I ask near the end of this comment @jgerigmeyer: https://github.com/oddbird/oddleventy/pull/205#pullrequestreview-1057833173

mirisuzanne: We could

1. find a way to render custom-properties in herman color swatches. I think we have some of the infrastructure needed to do that already in sizes?
2. render the map here instead of the color-swatches by creating an 'example' that outputs the result of `meta.module-variables()` (the map that Sass generates for us)

Option 1 is clearly the better solution, if someone has the time to implement it in Herman?

jgerigmeyer: Maybe file an issue (or trello card) in Herman, then go with option 2 for now? We could render colors in an iframe with custom CSS like we do for sizes/ratios/icons/examples/fonts.

stacyk: New card created for the longer term fix: https://trello.com/c/6W6pXJtv/93-render-colors-in-herman-in-an-iframe-with-custom-css-like-we-do-for-sizes-ratios-icons-examples-fonts

@mirisuzanne could you help me with 2 above? I'm not sure exactly what this looks like.

mirisuzanne: @stacyk Pushed an example with content-colors

src/scss/config/color/_index.scss
```diff
@@ -25,6 +25,21 @@
...
+///   @use 'config/color/content';
+///   @use 'config/tools';
+///   @use 'sass:meta';
+///
+///   @each $name, $color in tools.compile-colors(meta.module-variables('content')) {
+///   /*! #{$name}: #{$color}; */
+///   }
 @include herman.add(
   'colors',
   'content-colors',
```
stacyk, line 45: @mirisuzanne doesn't this `content-colors` here reference the `/// @colors content-colors` line that we no longer have from the `src/scss/config/color/_content.scss` file?

mirisuzanne: uh, yes, I guess so - that line is adding the colors to Herman, even though (for now) we're not displaying them with Herman. So it's not doing anything useful, but also not causing any issues. We could remove that `herman.add` for now, or just leave it so it works once we add custom property support to Herman? Either way.

stacyk: OK, I'd say we can leave it there for now since it doesn't hurt anything and maybe if/when we update Herman we will be in good shape.

stacyk (approved):
ok @mirisuzanne @jgerigmeyer this has all the colors showing values instead of the blank previews (although the /*! is a bit messy but I think that is so it doesn't error). This will work until we have a better fix in place for Herman.

Screen Shot 2022-08-11 at 11 46 57 AM

The ordering of things on this page doesn't bother me too much but I think it we want a certain order (like having the h1 at the top) we need different pages and I think I prefer them all one one page more than my desire to reorder at this time.

jgerigmeyer (approved):

mirisuzanne:
@stacyk The `/*! comment */` syntax come from Sass, and I just default to using it when we want to see the comment in the output. It seems like maybe we don't need that, since we're not doing compressed output? https://sass-lang.com/documentation/syntax/comments#in-scss

stacyk:
> @stacyk The `/*! comment */` syntax come from Sass, and I just default to using it when we want to see the comment in the output. It seems like maybe we don't need that, since we're not doing compressed output? https://sass-lang.com/documentation/syntax/comments#in-scss

OK, I updated the comment (removed the `!`). Is this ready to merge now @mirisuzanne?

mirisuzanne (approved):
looks great!

PR #206, opened 2022-04-22
Click to copy header anchor links

sanajaved7:
## Steps to test/reproduce

## Show me


jgerigmeyer:
[As a reader, I love "click to copy" an anchor link to Headlines (IDs).](https://trello.com/c/FCzEvLWw/79-as-a-reader-i-love-click-to-copy-an-anchor-link-to-headlines-ids)

src/filters/type.js
```diff
@@ -16,7 +16,18 @@ const mdown = markdown({
...
-  .use(mdAnchor)
+  .use(mdAnchor, {
+    level: [2],
+    permalink: mdAnchor.permalink.linkAfterHeader({
+      symbol:
+        '<svg data-icon="link" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false" viewBox="0 0 20 20"> <path d="M10.682 12.998c-0.943 0-1.886-0.359-2.604-1.077-0.195-0.195-0.195-0.512 0-0.707s0.512-0.195 0.707 0c1.046 1.046 2.747 1.046 3.793 0l3.636-3.636c1.046-1.046 1.046-2.747 0-3.793s-2.747-1.046-3.793 0l-3.068 3.068c-0.195 0.195-0.512 0.195-0.707 0s-0.195-0.512 0-0.707l3.068-3.068c1.436-1.436 3.772-1.436 5.207 0s1.436 3.772 0 5.207l-3.636 3.636c-0.718 0.718-1.661 1.077-2.604 1.077z"></path><path d="M4.682 18.998c-0.943 0-1.886-0.359-2.604-1.077-1.436-1.436-1.436-3.772 0-5.207l3.636-3.636c1.436-1.436 3.772-1.436 5.207 0 0.195 0.195 0.195 0.512 0 0.707s-0.512 0.195-0.707 0c-1.046-1.046-2.747-1.046-3.793 0l-3.636 3.636c-1.046 1.046-1.046 2.747 0 3.793s2.747 1.046 3.793 0l3.068-3.068c0.195-0.195 0.512-0.195 0.707 0s0.195 0.512 0 0.707l-3.068 3.068c-0.718 0.718-1.661 1.077-2.604 1.077z"></path></svg>',
+      style: 'visually-hidden',
+      assistiveText: (title) => `Permalink to “${title}”`,
+      visuallyHiddenClass: 'visually-hidden',
```
jgerigmeyer, line 26:
```suggestion
      visuallyHiddenClass: 'sr-only',
```

src/scss/layout/_anchor-link.scss
```diff
@@ -0,0 +1,30 @@
+@use '~accoutrement/sass/tools';
+
+.visually-hidden {
+  display: none;
+}
+
```
jgerigmeyer, line 6:
```suggestion
```

src/js/clickToCopy.js
```diff
@@ -0,0 +1,25 @@
...
+
+const clickToCopy = function () {
+  const anchorLinks = document.querySelectorAll('.header-anchor');
+
+  anchorLinks.forEach((anchorLink) =>
+    anchorLink.addEventListener('click', () => copyAnchorLink(anchorLink)),
+  );
+};
+
+module.exports = clickToCopy;
```
jgerigmeyer, line 25: This is the first runtime JS we have on the site, so I think it'd be good to add tests (and test coverage) for it. I think that would involve a few changes to the `jest.config.js` -- or possibly separate config files for client-tests and src-tests or something. I'd be happy to do that or pair with you on it, but also happy if you want to go ahead and tackle it.

test/js/type.test.js
```diff
@@ -14,7 +16,7 @@ describe('typography filters', () => {
...
...wrapper">\n<h2 id="${slugifyd}" tabindex="-1">${typogrd}</h2>\n<a class="header-anchor" href="#lorem-ipsum-dolor-sit-amet%2C-consectetur"><span aria-hidden="true"><svg data-icon="link" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false" viewBox="0 0 20 20"> <path d="M10.682 12.998c-0.943 0-1.886-0.359-2.604-1.077-0.195-0.195-0.195-0.512 0-0.707s0.512-0.195 0.707 0c1.046 1.046 2.747 1.046 3.793 0l3.636-3.636c1.046-1.046 1.046-2.747 0-3.793s-2.747-1.046-3.793 0l-3.068 3.068c-0.195 0.195-0.512 0.195-0.707 0s-0.195-0.512 0-0.707l3.068-3.068c1.436-1.436 3.772-1.436 5.207 0s1.436 3.772 0 5.207l-3.636 3.636c-0.718 0.718-1.661 1.077-2.604 1.077z"></path><path d="M4.682 18.998c-0.943 0-1.886-0.359-2.604-1.077-1.436-1.436-1.436-3.772 0-5.207l3.636-3.636c1.436-1.436 3.772-1.436 5.207 0 0.195 0.195 0.195 0.512 0 0.707s-0.512 0.195-0.707 0c-1.046-1.046-2.747-1.046-3.793 0l-3.636 3.636c-1.046 1.046-1.046 2.747 0 3.793s2.747 1.046 3.793 0l3.068-3.068c0.195-0.195 0.512-0.195 0.707 0s0.195 0.512 0 0.707l-3.068 3.068c-0.718 0.718-1.661 1.077-2.604 1.077z"></path></svg></span> <span class="visually-hidden">Permalink to “Lorem ipsum dolor sit amet, consectetur”</span></a></div>\n`;
```
jgerigmeyer, line 19: This seems a bit fragile to me. I'd prefer either pulling out this shared SVG markup into its own file (imported both here and in the `type.js` file), and/or adjusting this test to just check that the markup contains a few specific things (instead of comparing the entire exact string).

jgerigmeyer (commented):
Looks great! Functionally I had three thoughts:

1. I think it'd be cool if we could add some animation when the anchor is clicked.
2. What about setting a `setTimeout` so the `Link Copied` text disappears again after a few seconds?
3. Using it, I was surprised that the anchor link was also followed -- which caused a page jump. Personally I think if we're doing click-to-copy we could add a `preventDefault` so that the link click doesn't jump the page. 🤷

test/js/type.test.js
sanajaved7: Agreed, good catch! I'll rework this a bit, thanks!

src/js/clickToCopy.js
sanajaved7: Agree 100%! I'm happy to take a stab at it and reach out if I get stuck! 🙏🏼

src/js/clickToCopy.js
```diff
@@ -0,0 +1,35 @@
...
+const copyAnchorLink = function (anchorLink) {
+  navigator.clipboard.writeText(anchorLink.href).then(() => {
+    removeAnchorClickedClass();
+    anchorLink.classList.add('anchor-clicked');
+
+    setTimeout(() => {
+      anchorLink.classList.remove('anchor-clicked');
+    }, 3000);
+  });
+};
```
jgerigmeyer, line 22: Minor point, but I think we could rewrite this using async/await, which I personally find to be more readable:

```suggestion
const copyAnchorLink = function (anchorLink) {
  navigator.clipboard.writeText(anchorLink.href).then(() => {
    removeAnchorClickedClass();
    anchorLink.classList.add('anchor-clicked');

    setTimeout(() => {
      anchorLink.classList.remove('anchor-clicked');
    }, 3000);
  });
};
```

It looks like any browsers that support the `navigator.clipboard.writeText()` async functionality also support `async/await`.

Should we test out browser support for `navigator.clipboard.writeText()`, or consider a polyfill/fallback of some sort?

src/js/clickToCopy.js
```diff
@@ -0,0 +1,35 @@
...
+
+const copyAnchorLink = function (anchorLink) {
+  navigator.clipboard.writeText(anchorLink.href).then(() => {
+    removeAnchorClickedClass();
+    anchorLink.classList.add('anchor-clicked');
+
+    setTimeout(() => {
+      anchorLink.classList.remove('anchor-clicked');
+    }, 3000);
```
jgerigmeyer, line 20: Very minor (maybe not even worth addressing), but I note in practice that since we're never canceling this `setTimeout` call on a subsequent click within 3 sec, it's possible to click a link, click it again 2 seconds later, then the class is removed right away after the second click.

If we wanted to address it, I think we could save the id and then use `clearTimeout` to clear it if it gets clicked again within 3 seconds. 🤷‍♂️

test/js/clickToCopy.test.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * @jest-environment jsdom
+ */
+/* eslint-disable no-undef */
```
jgerigmeyer, line 4:
```suggestion

/* eslint-env browser */

```

test/js/clickToCopy.test.js
```diff
@@ -0,0 +1,51 @@
...
+    const evt = new Event('click');
+    const anchorLink = document.querySelector('.header-anchor');
+    anchorLink.dispatchEvent(evt);
+
+    expect(window.navigator.clipboard.writeText).toHaveBeenCalledWith(
+      anchorLink.href,
+    );
+  });
+
+  test('copyAnchorLink adds and removed clicked class', async () => {
```
jgerigmeyer, line 34:
```suggestion
  test('copyAnchorLink adds and removes clicked class', async () => {
```

jgerigmeyer (commented):
Functionality looks good! Left some minor code comments, but I think this is really close. I'd still love an animation personally 😄

test/js/clickToCopy.test.js
```diff
@@ -0,0 +1,51 @@
...
+  beforeAll(() => {
+    Object.assign(window.navigator, {
+      clipboard: {
+        writeText: jest.fn().mockImplementation(() => Promise.resolve()),
+      },
+    });
+    // eslint-disable-next-line max-len
+    const anchorLinkHTML = `<div class="anchor-link-wrapper"><h2 id="background" tabindex="-1">Background</h2><a class="header-anchor" href="#background"><span aria-hidden="true">${anchorLinkIconString}</span> <span class="sr-only">Click to copy the permalink to “Background”</span></a></div>`;
+    document.body.innerHTML = anchorLinkHTML;
+  });
```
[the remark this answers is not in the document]
sanajaved7: @jgerigmeyer I think if there was some data cleanup or other shared wrap-up functionality I would definitely want an `afterAll` here but I don't have anything the tests do or share after the assertions that could go there.

src/js/clickToCopy.js
sanajaved7: @jgerigmeyer that's a great point about the browser support for the navigator clipboard API - a quick search shows that it will work with [most](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard#browser_compatibility) browsers including Edge, but not IE. If we wanted to have a fallback for IE, we could explore using `Document.execCommand()` but, that has been deprecated. Do you want me to do some more research for IE support?

jgerigmeyer: I'm not too concerned about support for IE, but if there's a way we can make it not broken in IE that'd be great -- for example, checking browser support for the clipboard api, and just allowing the click to flow through normally on older browsers.

test/js/clickToCopy.test.js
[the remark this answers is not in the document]
jgerigmeyer: My thinking was that since were mocking a global `window.navigator` object, any other tests or code that also interact with that object would be impacted by this. In this case I don't think that's a practical problem, so it's all a bit theoretical. But I might lean toward something like:

```suggestion
  let clipboard;

  beforeAll(() => {
    clipboard = window.navigator.clipboard;
    window.navigator.clipboard = {
      writeText: jest.fn().mockResolvedValue(),
    };
    // eslint-disable-next-line max-len
    const anchorLinkHTML = `<div class="anchor-link-wrapper"><h2 id="background" tabindex="-1">Background</h2><a class="header-anchor" href="#background"><span aria-hidden="true">${anchorLinkIconString}</span> <span class="sr-only">Click to copy the permalink to “Background”</span></a></div>`;
    document.body.innerHTML = anchorLinkHTML;
  });

  afterAll(() => {
    window.navigator.clipboard = clipboard;
  });
```

src/js/clickToCopy.js
sanajaved7: @jgerigmeyer I think we were hoping that if the click triggers the text to display and users see "Link Copied" they likely won't click the link again or need to (especially within 3 seconds 😅) but if that seems weird, I'm happy to refine this a bit more.

src/js/clickToCopy.js
```diff
@@ -0,0 +1,39 @@
...
+      removeAnchorClickedClass();
+      anchorLink.classList.add('anchor-clicked');
+
+      setTimeout(() => {
+        anchorLink.classList.remove('anchor-clicked');
+      }, 3000);
+    });
+  } catch (error) {
+    throw new Error('An error occurred with copying the anchor link.');
+  }
```
jgerigmeyer, line 25: This is fine, but I'm not sure re-throwing a custom error is practically any different than just allowing the original error to be thrown?

If we want to do this, I think the purpose is to *not* throw an error, and also not to hijack the click. So in that case maybe better to wrap the `copyAnchorLink(anchorLink);` and `e.preventDefault();` in a try-catch block, and just swallow the error entirely?

But really, this seems pretty unimportant. It looks like we have way bigger issues in IE11 anyway 😅 

Screen Shot 2022-05-13 at 10 30 12 AM

jgerigmeyer (approved):

src/scss/layout/_anchor-link.scss
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+
+  &:visited {
+    --fill: var(--active);
+  }
+}
+
+.anchor-clicked::before {
+  background: var(--feature);
```
stacyk, line 43: I would add an animation in and out on the text that is shown when copied. ... something like 

``` 
@keyframes fadein-out {
  10% {
    opacity: 1;
  }

  90% {
    opacity: 1;
  }
}

.anchor-clicked::before {
animation: fadein-out 3000ms forwards;
...
opacity: 0; // so it is not shown right away
```

Also we typically will nest the `::before`. It helps if you are adding other styles to `.anchor-clicked` like:

```
.anchor-clicked {
  &::before {
  ...
  }
  
  svg {
    transform: rotate(-45deg);
  }
}  
```

src/scss/layout/_anchor-link.scss
```diff
@@ -0,0 +1,59 @@
...
+    --anchor-position-offset-t: 1.25em;
+  }
+
+  &.anchor-clicked {
+    @include tools.transition('move');
+
+    svg {
+      transform: rotate(-45deg);
+    }
+  }
```
stacyk, line 35: @dvdherron 

```suggestion
  svg {
    @include tools.transition('move');
  }
```

and then I'd move the svg transform to the .anchor-clicked (see comment a few steps belore that refers to the animation.)

```
.anchor-clicked {
...

  svg {
    transform: rotate(-45deg);
  }
...

```

src/scss/layout/_anchor-link.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.anchor-clicked::before {
+  background: var(--feature);
+  border-radius: var(--radius);
+  color: var(--bg);
+  content: 'Link Copied';
+  font-weight: bold;
+  font-size: var(--xxsmall);
```
stacyk, line 48:
```suggestion
  font-size: var(--xxsmall);
  font-weight: bold;
```

src/scss/layout/_anchor-link.scss
```diff
@@ -0,0 +1,59 @@
...
+  left: var(--anchor-text-offset-l, -0.25em);
+  padding: var(--quarter-shim) var(--half-shim);
+  position: absolute;
+  top: var(--anchor-text-offset-t, -2.25em);
+  white-space: nowrap;
+
+  @include tools.above('wide-break') {
+    --anchor-text-offset-l: -6.5em;
+    --anchor-text-offset-t: 0.5em;
+  }
```
stacyk, line 58: @dvdherron there are a number of numbers 😵‍💫 used in this file. Are any related? I see two that are -.25em ... would it make sense to make a few variables?

stacyk (commented):
Looking good so far and I am excited to see this one coming together. I left a few suggestions that I hope make sense.

src/scss/layout/_anchor-link.scss
```diff
@@ -0,0 +1,59 @@
...
+    @include tools.transition('move');
+
+    svg {
+      transform: rotate(-45deg);
+    }
+  }
+
+  &:visited {
+    --fill: var(--active);
+  }
```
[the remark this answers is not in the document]
SondraE: @dvdherron @stacyk Yeah, a different visited color or state could be nice. Maybe the visited color is a new purple we define?? Or maybe the visited indication is that the icon remains vertical. (I'm not seeing a visited color in Safari, just in Chrome. Is that expected?)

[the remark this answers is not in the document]
dvdherron: I changed this to where the `:visited` state is more apparent. 

Light mode, unvisited
image

Hover:
image

And visited is a bit lighter:

image

@stacyk Any idea why `visited` would work or not in different browsers? The above works fine in FF but I can't get Chrome to show the correct visited state.

src/scss/layout/_anchor-link.scss
dvdherron: good catch. I also added an animation to the `link copied` text

src/scss/layout/_anchor-link.scss
dvdherron: fixed

src/scss/layout/_anchor-link.scss
dvdherron: adjusted

src/scss/layout/_anchor-link.scss
dvdherron: @stacyk  I tried to clean this up by making some variables and simplifying what happens on widescreens. I think it works fine if "Link Copied" stays on top of the icon no matter the screen width?

src/filters/type.js
```diff
@@ -10,13 +10,25 @@ const striptags = require('striptags');
...
   typographer: true,
 })
   .disable('code')
-  .use(mdAnchor)
+  .use(mdAnchor, {
+    level: [2],
+    permalink: mdAnchor.permalink.linkAfterHeader({
+      symbol: anchorLinkIconString,
+      style: 'visually-hidden',
+      assistiveText: (title) => `Click to copy the permalink to “${title}”`,
```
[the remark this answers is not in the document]
dvdherron: hmmm. I think removing "click to" is a good call. Having "copy the permalink to ..." first makes it clearer from the start what pressing this link will do so I think I would vote for keeping the title towards the end.

stacyk (commented):
Just a few tiny touchups left @dvdherron 

@sanajaved7 @dvdherron does anyone know why on a small screen/phone, if you visit the pasted/anchor link the position of the content is touching the left viewport and am scrolled down a bit?

For instance, visit this link on my phone and I see:

https://user-images.githubusercontent.com/1581694/173150855-0ea14f4d-eb44-4f7c-a933-84cadd692e1c.mov

Maybe it is because of that Netlify banner?

src/scss/layout/_anchor-link.scss
```diff
@@ -0,0 +1,67 @@
...
+    opacity: 0;
+    padding: var(--quarter-shim) var(--half-shim);
+    position: absolute;
+    top: calc(var(--top-position-offset-default) * -1);
+    white-space: nowrap;
+  }
+
+  svg {
+    transform: rotate(-45deg);
+    transition: tools.change('move');
```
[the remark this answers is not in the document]
sanajaved7: @stacyk @dvdherron When running the changes locally, I am seeing the same issue (content touching left viewport) so I don't think its because of the Netlify banner. I did notice there is a `overflow-x: hidden` style and when I toggle that, the viewport seems to be correct - could that be the culprit? 
Screen Shot 2022-06-13 at 5 13 20 PM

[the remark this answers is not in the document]
dvdherron: @sanajaved7 @stacyk Hmm. I've this happen sometimes, but it doesn't happen consistently. . . I'll take another look to see if it's something we've done here that's causing it.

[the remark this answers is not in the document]
dvdherron: and @stacyk I moved the transition to `header-anchor`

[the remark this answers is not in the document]
stacyk: @dvdherron when I turn my phone to get the close icon on that netlify banner, I am not seeing the same issue anymore so I am guessing we won't have this same problem when we test this on the live site.

stacyk (approved):
Can't wait to get this posted! Nice job @sanajaved7 @dvdherron!

src/scss/layout/_anchor-link.scss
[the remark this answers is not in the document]
stacyk: oops, I just read @sanajaved7's message again about testing locally ... I still this we can move forward with this and create a new card for the small screen anchor-landing positioning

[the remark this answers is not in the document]
dvdherron: @stacyk @sanajaved7 Okay, after playing with this some more I think I've found a way where I can consistently make this overflow happen. I think it has something to do with the length of the section titles on each page. 

- Compare the ["Dependabot"](https://deploy-preview-206--oddleventy.netlify.app/2022/06/01/dependabot-single-pull-request/) article (longer headings) with, say, the "Support (Not) Unknown" article. 
- If you click and copy all the headings on the "Support" article, the overflow never happens. 
- If you click the longer heading in the Dependapot article ("Upgrade Node (JavaScript) Dependencies") the overflow happens and, depending on the browser, the page gets stuck in the overflow view and I think you have to reload the original link to the article to correct the layout. 

Can you all duplicate this?

src/filters/type.js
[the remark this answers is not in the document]
dvdherron: Assistive text changed in https://github.com/oddbird/oddleventy/pull/206/commits/093a3ab3b996a56d524c14401677a48cdc337eb5

SondraE (approved):
LGTM! 
The experience is a bit different in each browser I tested (Chrome, Firefox, and Safari), but it sounds like maybe those difference are going to be worked out in new cards.

PR #219, opened 2022-07-11
Upgrade deps.

jgerigmeyer:


jerivas (approved):
Seems it's mostly renaming js to cjs. Why do we want to do that?

PR #220, opened 2022-07-11
Remove markdown from u-url links

jgerigmeyer:
https://oddbird.slack.com/archives/C02HLQVD1/p1657569770885699

### DIFF
--- a/content/_includes/page/base.njk
+++ b/content/_includes/page/base.njk
@@ -26,7 +26,7 @@
   </div>
 
   <!-- used for webmentions -->
-  <a href="{{ page.url | url }}" class="u-url" hidden>{{ (banner or title) | mdInline | safe }}</a>
+  <a href="{{ page.url | url }}" class="u-url" hidden>{{ (banner or title) | removeMd }}</a>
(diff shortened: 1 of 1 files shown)


stacyk (commented):
@jgerigmeyer is it supposed to include the asterisks in the code in the webmentions line on the homepage? `Custom Web App *Design, Development, & Training*`

Screen Shot 2022-07-11 at 3 53 21 PM

jgerigmeyer:
> @jgerigmeyer is it supposed to include the asterisks in the code in the webmentions line on the homepage? `Custom Web App *Design, Development, & Training*`

@stacyk Yeah, I noticed that too -- and I'm not sure why the [remove-markdown](https://github.com/tommoor/remove-markdown) tool isn't removing those. But from what I understand, the contents of the `.u-url` tag doesn't actually matter at all, so I'm not sure it's worth messing with? In fact, the pattern I see elsewhere seems to be just having the `href` as the contents as well -- maybe we should do that?

stacyk (approved):

PR #221, opened 2022-08-02
Initial draft of zero-units blog post

mirisuzanne:
## Description

A blog post about removing units from zeros in CSS.

todo:
- [ ] hero image

## Steps to test/reproduce

- /2022/08/04/zero-units/


mirisuzanne:
[Zero Units Blog Post](https://trello.com/c/wKCYmPMX/183-zero-units-blog-post)

jgerigmeyer (approved):

jgerigmeyer (approved):

content/blog/2022/zero-units.md
```diff
@@ -0,0 +1,348 @@
...
+I do that with a `calc()` function:
+
+```css
+ol, ul {
+  /* (1em + 0 == 1em) and (1em + -1em == 0) */
+  padding-inline-start: calc(1em + var(--outdent));
+}
+
+li {
+  /* nested lists should not outdent */
```
stacyk, line 89: Is this comment correct? should it be a nested list instead of an li?

stacyk (commented):
I only got half way through so far because I couldn't replicate the demo as written ... I'll finish later today.

content/blog/2022/zero-units.md
```diff
@@ -0,0 +1,348 @@
...
+  --outdent: 0;
+}
+```
+
+There would be other ways to do this of course --
+but it made sense to me
+as a way of handling different outdent styles
+with a single variable toggle.
+
+**Sadly, the code above doesn't work.**
```
stacyk, line 99: bwahahahahhaha... I just thought oh wait, this whole article is about things that don't work.

content/blog/2022/zero-units.md
stacyk: ah, I see... I noticed I was setting the --outdent value in the ol, ul which is why when I added this here, it didn't do anything (event with the unit). When I moved the outdent value to the `html {}` then the `.ol, .ul` didn't override this li anymore.

content/blog/2022/zero-units.md
```diff
@@ -0,0 +1,348 @@
...
+I had a similar issue last week,
+with a CSS minifier
+_removing all Cascade Layers from my CSS_.
+In one case,
+the transformation is an over-eager 'best practice'.
+In the other it's an over-eager attempt
+to remove 'unknown syntax'.
+In both cases,
+_I wish linters and minifiers
+would be less eager to transform my code_.
```
stacyk, line 337: 👍

content/blog/2022/zero-units.md
```diff
@@ -0,0 +1,348 @@
...
+
+I think we (as an industry)
+tend to adopt rules and 'best practices' very quickly,
+without communicating the caveats clearly,
+and then we fail to update our understanding as things change.
+Tools based on these 'best practice' trends
```
stacyk, line 343: Earlier in this article "best practices" was in Title Case, but this is all lowercase. Not sure if there is a reason to switch when it's written here?

stacyk (approved):
This is a great explanation to something I don't see much talk about. Loved it.

content/blog/2022/zero-units.md
mirisuzanne: oh right, you also need the outer setting - I can add that to the demo. I use `li` to change it for nested lists - since anything nested would be inside an `li`.

content/blog/2022/zero-units.md
```diff
@@ -0,0 +1,359 @@
...
+for raw zero values,
+directly applied to properties like
+`margin` or `padding` --
+but there are other places
+where this 'best practice' will break your CSS.
+
+In general:
+_when zero is inside a function, the type of zero matters_.
+(At least,
+that's where I've always encountered the issue.)
```
stacyk, line 224: It might be interesting to close the article with a audience call to action, like what cases have you come across ...   or maybe as part of the tweet that shares this @mirisuzanne

PR #222, opened 2022-08-02
Replace Google Analytics with Plausible

jerivas:


jerivas:
[Replace Google Analytics with Plausible Analytics.](https://trello.com/c/exkGY78f/86-replace-google-analytics-with-plausible-analytics)

content/_layouts/base.njk
```diff
@@ -59,16 +59,8 @@
...
-      <script async src="https://www.googletagmanager.com/gtag/js?id={{ site.google_analytics_id }}"></script>
-      <script>
-        window.dataLayer = window.dataLayer || [];
-        function gtag(){dataLayer.push(arguments);}
-        gtag('js', new Date());
-
-        gtag('config', '{{ site.google_analytics_id }}');
-      </script>
+    {% if env.ELEVENTY_ENV === 'production' -%}
+      <script defer data-domain="oddbird.net" src="https://plausible.io/js/plausible.js"></script>
```
jerivas, line 63: Looks like this enables tracking on PR previews as well, which is convenient, but maybe it muddies the real tracking data?

jgerigmeyer: Hm, interesting. We might want to use the `CONTEXT` [env var](https://docs.netlify.com/configure-builds/environment-variables/#build-metadata) and [pass that through](https://docs.netlify.com/integrations/frameworks/environment-variables/#netlify-variables) to our 11ty build command to opt out for deploy previews? I don't think we want to include them (or local dev) unless there's a quick way to filter them out.

jerivas: I changed the conditional to read from `CONTEXT` in 82c1b81e8d4b103f2d59f74616969a6b55da2d43 and it seems to be working as expected. This was also the only instance of `ELEVENTY_ENV` in the whole codebase so we can go ahead and remove it, unless it's used by Eleventy itself 🤷‍♂️

stacyk (commented):
LGTM!

stacyk (approved):

jgerigmeyer (approved):

PR #223, opened 2022-08-02
quotes

SondraE:
photo-1504618223053-559bdef9dd5a

## Steps to test/reproduce
- See new quotes on Work list page
- See new quotes on Metecho and MetaDeploy detail pages
- Second paragraph at the top of /services/development/ (removed word "modern")


SondraE:
[Add Salesforce quote from David Reed.](https://trello.com/c/tlxvkxQr/173-add-salesforce-quote-from-david-reed)

SondraE:
[Services Page > Web Design & Dev section: "modern" adjective in front of Sass/CSS feels like it implies we write non-modern code in other languages](https://trello.com/c/rFRrGHGE/116-services-page-web-design-dev-section-modern-adjective-in-front-of-sass-css-feels-like-it-implies-we-write-non-modern-code-in-oth)

jgerigmeyer (approved):

stacyk (approved):
@SondraE since we have a few quotes from Salesforce, would it make sense to show say, Jason's quote on the work Sample page and then once you click into the Metecho Case Study page, maybe only show David and Brandon's quotes?  Currently we see Jason's same quote on the work Sample list page and as the first quote on the Metecho page.

Screen Shot 2022-08-08 at 2 05 25 PM

Same goes for MetaDeploy. Currently we see David Reed's quote on the work sample page, then it is also the first quote on the single. Then it is a different Jason quote (MD specific)  followed by the same two Brandon and Jason quotes we saw on Metecho. Would it make sense to show Jason's new quote and maybe Brandon's quote here on the MD case study page and remove the other two that were used on the work sample page?

I realize people may not go tot he work sample page before they go into a case study, but it feels like for these projects we have enough good quotes to spread them out. I am approving as I know this suggestion may not go with the strategy of the site so feel free to merge.

SondraE:
@stacyk Good questions!

Currently, our pattern on all clients (except TimeDesigner, I see) is to repeat the same quote on the list page and the detail page whether there's one or multiple quotes from the client. It makes some sense to me to do this since people may come directly to a client detail page without seeing the list page (like you mentioned). Should we reconsider this pattern across clients?

I'm on board with your suggestion not to repeat quotes between the projects. I'll make that change.

PR #226, opened 2022-08-16
Add talks and podcasts

mirisuzanne:
## Description

talks and podcasts: 
- /2022/05/30/word-wrap-17/
- /2022/08/16/podrocket-s2/
- /talks/cascading-layers/
- /talks/queries-units/


mirisuzanne:
[PodRocket](https://trello.com/c/kO9vOK2m/94-podrocket)

mirisuzanne:
[Word Wrap Show](https://trello.com/c/S5BLe770/86-word-wrap-show)

jgerigmeyer (approved):

content/talks/cascading-layers.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Cascading Layers of !mportance
+date: 2022-08-16
```
jgerigmeyer, line 3: I realized that on some talks we date the file to the talk date, whereas others we date to the announcement date. 🤷

mirisuzanne: @jgerigmeyer I think some of the talks were added after they happened, and we backdated the page date?

But it's also come up several times that the way we handle talks is a bit confusing on the resource page, and in the RSS feed: 

- https://trello.com/c/aesmWWbl/85-i-want-upcoming-events-to-not-take-over-the-top-of-the-resources-list-page (I'm not sure how that was resolved, since there's no discussion, and no linked PR)
- https://trello.com/c/C4AKsJpG/125-its-confusing-that-talks-are-listed-at-their-most-recent-date-but-show-the-created-date-on-the-resources-page (here, I think we removed the page date from the card)
- https://trello.com/c/Tl0dY6xA/185-rss-feed-output-isnt-very-useful-for-talk-announcements (I just opened this issue the other day, because of this merge)

Maybe we need to overall rethink and document how talks are handled?

jgerigmeyer: Sure, that makes sense. I think that first one is the one I was noticing -- I'll move that back into an active list.

PR #227, opened 2022-08-16
Draft container query synax article

mirisuzanne:
## Description

an article about using the most up-to-date syntax for container queries

- /2022/08/16/cq-syntax/

**Remember to update the pub date before merging**


jgerigmeyer (approved):

content/blog/2022/cq-syntax.md
```diff
@@ -0,0 +1,365 @@
+---
+title: Use the Right Container Query Syntax
+sub: Size queries are stable, and shipping in browsers this summer
```
stacyk, line 3: @mirisuzanne Should we use a non relative time here since this article may be referenced for a while?

content/blog/2022/cq-syntax.md
```diff
@@ -0,0 +1,365 @@
...
+The downside is,
+many people (including me!) have written articles,
+made demos,
+given talks,
+and released videos
+using now-out-of-date syntax.
+
+So, what syntax
+is actually shipping in browsers
+this summer?
```
stacyk, line 69: @mirisuzanne  Again, the "this summer" might be off when someone reads this next year, and I know this stuff might be old news to us by then but I am sure some people won't start researching this until they have to do their next redesign/project and that may be a while from now.

content/blog/2022/cq-syntax.md
```diff
@@ -0,0 +1,365 @@
...
+
+## TL;DR, a code sample
+
+```css
+main, .sidebar {
+  /* shorthand: `container: page-layout / inline-size;` */
+  container-type: inline-size;
+  container-name: page-layout;
```
stacyk, line 77: @mirisuzanne Curious, does the meaning get lost if we order these by alpha? That is also how the shorthand is ordered as well so it may make sense here too.

content/blog/2022/cq-syntax.md
```diff
@@ -0,0 +1,365 @@
...
+about a given container.
+
+The initial value is `normal`.
+By default,
+all elements are _style containers_ --
+meaning we should be able to query the
+[computed value](https://developer.mozilla.org/en-US/docs/Web/CSS/computed_value)
+of any property on any element.
+But don't worry about that yet,
+**browsers are only shipping size queries this summer**.[^style]
```
stacyk, line 125: If you decide to change this above, you may want to update it here as well.

content/blog/2022/cq-syntax.md
```diff
@@ -0,0 +1,365 @@
...
+
+Note that _names go first_, before the slash.
+**This is one of the big changes
+that will break a lot of demos.**
+In an earlier draft of the feature,
+the order was reversed.
+If you see an old container query
+article or talk or demo
+that no longer works,
+_this is probably why_.
```
stacyk, line 245: Do we have a callout for this kind of stuff? It feels like this is what people will want to skim to if they are looking for the change

content/blog/2022/cq-syntax.md
```diff
@@ -0,0 +1,365 @@
...
+```css
+@container not (inline-size > 40em) { /* … */ }
+```
+
+Safari currently reads that as
+a container with the name of `not`,
+and an `inline-size` that is _greater-than_ `40em`.
+You can get the correct behavior
+in all browsers
+by wrapping the entire query in parentheses:
```
stacyk, line 361: interesting!

content/blog/2022/cq-syntax.md
```diff
@@ -0,0 +1,365 @@
...
+  [size container features](https://drafts.csswg.org/css-contain-3/#size-container)
+  that can be queried --
+  like `aspect-ratio` and `orientation`.
+
+## Browser support
+
+Container size queries & units
+are expected to ship
+in both Chrome & Safari
+sometime this summer.
```
stacyk, line 330: there it is again :)

stacyk (approved):
@mirisuzanne I left a few suggestions but overall this would be good to go either way! I think this article is going to be a really good resource for people to use. I wonder if we should note that we will keep it updated incase there are future changes or additions, etc.

content/blog/2022/cq-syntax.md
mirisuzanne: We don't yet. We could maybe use something similar to the callouts that David added to css-sandbox? See https://css.oddbird.net/sass/color-spaces/proposal/ for example - with both a 'note' and a 'warning' version of the callout. @SondraE maybe you want to weigh in on that?

SondraE: @mirisuzanne @stacyk Great idea to add a design element to highlight that info! Makes sense to me to use what David used. Or maybe with slight adjustments in this direction. https://zenplates.co/patterns/callouts

mirisuzanne:
@stacyk updated, and ready for re-review. (I already rand the callout styles past Sondra, and will open a trello card for styling more callout `types` - since we currently don't have UI colors for things like errors or warnings.

content/blog/2022/cq-syntax.md
```diff
@@ -0,0 +1,396 @@
...
+about a given container.
+
+The initial value is `normal`.
+By default,
+all elements are _style containers_ --
+meaning we should be able to query the
+[computed value](https://developer.mozilla.org/en-US/docs/Web/CSS/computed_value)
+of any property on any element.
+But don't worry about that yet,
+**browsers are only shipping size queries in September 2022**.[^style]
```
jgerigmeyer, line 146: Everywhere else we mention "late August" or "August 30", so I found this a bit confusing.

jgerigmeyer (approved):

stacyk (approved):
nice updates!

content/blog/2022/cq-syntax.md
mirisuzanne: Oops, yeah - pushed a fix

PR #229, opened 2022-08-22
Upgrade deps.

jgerigmeyer:


src/scss/config/_scale.scss
```diff
@@ -75,7 +75,7 @@ $spacing-sizes: (
 /// @group config-scale
 $theme-sizes: (
   'page-margin': calc(var(--quarter-shim) + 4vw),
-  'border-width': 3px,
+  'border-width': medium,
```
jgerigmeyer, line 78: @stacyk This results in the updated HTML as expected, but the border itself is not visible.

Screen Shot 2022-08-22 at 12 28 03 PM

jgerigmeyer: @stacyk Looks good!

Screen Shot 2022-08-24 at 1 28 43 PM

jgerigmeyer (commented):
@stacyk Looks like I need approval here before merging...

stacyk (approved):
yay

PR #237, opened 2022-09-28
Upgrade deps and add script to automate.

jgerigmeyer:


package.json
```diff
@@ -34,7 +34,7 @@
     "site:clean": "rimraf './_site/!(assets)' './_site/assets/!(images)'",
     "assets:clean": "rimraf './_built/'",
     "serve": "npm-run-all build:assets build:site -p eleventy:serve watch:js watch:sass",
-    "test": "jest",
+    "test": "jest -i",
```
jgerigmeyer, line 37: There seems to be a bug with using multiple workers in Jest 29, so this limits it to one worker. I'll file an issue.

jerivas (approved):
Huh, didn't know we weren't running dep upgrades on this one 😅 LGTM 👍

package.json
jgerigmeyer: https://github.com/facebook/jest/issues/13332

PR #245, opened 2022-09-30
Use more explicit filters for hasData/withData.

jgerigmeyer:
Fixes article by David H showing up on David G's page.


jgerigmeyer:
[Review uses of `exact = false` in page filters.](https://trello.com/c/Ey8AVD5Y/191-review-uses-of-exact-false-in-page-filters)

PR #246, opened 2022-09-30
Fix shadow-repo workflow concurrency

jerivas:
Fixes https://github.com/oddbird/oddleventy/pull/241#pullrequestreview-1127214259

### DIFF
--- a/.github/workflows/shadow-repo.yml
+++ b/.github/workflows/shadow-repo.yml
@@ -8,7 +8,7 @@ on:
     types: [opened]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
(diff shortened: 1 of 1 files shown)


jerivas:
[Shadow repo workflows are being canceled](https://trello.com/c/VPLa71vW/193-shadow-repo-workflows-are-being-canceled)

jgerigmeyer:
@jerivas Do you think it would make sense to make this same change to the `test.yml` workflow?

.github/workflows/shadow-repo.yml
```diff
@@ -8,7 +8,7 @@ on:
     types: [opened]
```
jerivas, line 8: @jgerigmeyer I think we only need it here because the workflow is fired by a bunch of event types, and for each we do something different. `test.yml` only listens for a couple and they don't happen at the same time.

PR #247, opened 2022-09-30
Initial draft display of webmentions

mirisuzanne:
For the first version, we decided to punt on loading author images for eg the proposed 'facepile' of likes, or putting avatars next to author names.

Todo:
- [x] display the mentions
- [x] we might need to go through all the back mentions, and make sure they're curated? https://github.com/oddbird/oddleventy/pull/253

Right now we're also allowing mentions on our OSS tools, do we want that?

By default we're showing mentions with a type of `in-reply-to` (actual comments) or `mention-of` (which includes posts that just mention us). We could limit it to only the former, if we want. Or we could expand to include eg `bookmark-of` (similar to mentions) and/or `repost-of`.

## Steps to test/reproduce

- I imagine most mentions are on recent resources, such as the cq-syntax article


mirisuzanne:
[We should add webmentions](https://trello.com/c/xfmRD5Po/165-we-should-add-webmentions)

SondraE (approved):
I think this is looking good. 

Limiting to `in-reply-to` (actual comments) or `mention-of` seems best at least for now.

Will we need a regular schedule for reviewing and curating webmentions?

The facepile did help set the webmentions apart from the article. Perhaps when we add the Author Bios, we could put them at the end of the article above the webmentions to make it more immediately obvious that the webmentions are a different section/thing.

I think it's neat having webmentions on the OddTools. I like that Mia's mention of Cascading Colors on her site is linked on that Tool's page. The specific excerpt is a bit strange, but it did make me want to click through to Mia's site to see if that mention was an error or something, and I ended up reading some of the article.

mirisuzanne:
> Will we need a regular schedule for reviewing and curating webmentions?

Yes, I think that will be established as part of the automation that Ed is working on in a different story. 

> The facepile did help set the webmentions apart from the article. Perhaps when we add the Author Bios, we could put them at the end of the article above the webmentions to make it more immediately obvious that the webmentions are a different section/thing.

👍🏼

> I think it's neat having webmentions on the OddTools. I like that Mia's mention of Cascading Colors on her site is linked on that Tool's page. The specific excerpt is a bit strange, but it did make me want to click through to Mia's site to see if that mention was an error or something, and I ended up reading some of the article.

Right now for long mentions (over a certain length) we just show the first 250 characters of text. We could try to do something smarter with that, but I think that should be a new story.

stacyk:
Something is still a bit strange if you read through these messages since they come from different sources at different times so when you see your (@mirisuzanne) item, I don't know who or to which comment you are replying to: 
IMG_3150

I am also curious what makes one url truncate and the other run passed the container and then wrap?
urls

jgerigmeyer:
> Something is still a bit strange if you read through these messages since they come from different sources at different times so when you see your (@mirisuzanne) item, I don't know who or to which comment you are replying to

Agreed -- these need to be curated before we publish.

> I am also curious what makes one url truncate and the other run passed the container and then wrap?

@stacyk We're not doing the truncation there -- those URLs come directly from Twitter, which automatically truncates them in the webmention.

src/filters/mentions.cjs
```diff
@@ -1,92 +1,72 @@
...
+};
 
+const getTypes = (mentions, allow) =>
   // run all of the above for each webmention that targets the current URL
-  return mentions
-    .filter((entry) => allowedTypes.includes(entry['wm-property']))
+  mentions
+    .filter((entry) => (allow || defaultTypes).includes(entry['wm-property']))
     .filter(checkRequiredFields)
     .sort(orderByDate)
```
jgerigmeyer, line 66: @mirisuzanne If we're already sorting these in `forUrl`, do we need to sort them again here?

src/filters/mentions.cjs
```diff
@@ -1,92 +1,72 @@
...
-              Read more:
-              <a href="${entry['wm-source']}">${entry['wm-source']}</a>
-            </p>
-          `
-          : html,
-      );
-    } else {
-      entry.content.value = sanitizeHTML(text);
+  if (html) {
+    let truncated = truncate(html, 50, { byWords: true, ellipsis: '…' });
```
jgerigmeyer, line 39: @mirisuzanne This is now using a more consistent way to truncate the HTML, but it makes it difficult to know for sure when something has been truncated, so I removed the "Read more" addition. I think that's unnecessary anyway, when the title is already a link?

src/filters/mentions.cjs
```diff
@@ -1,92 +1,72 @@
...
-      Boolean(author) &&
-      Boolean(author.name) &&
-      Boolean(published) &&
-      Boolean(content)
-    );
-  };
+// only allow webmentions that have an author name and a timestamp
+const checkRequiredFields = (entry) => {
+  const { author, published, content } = entry;
+  return Boolean(author && author.name && published && content);
```
jgerigmeyer, line 58: @mirisuzanne If we're always filtering these out, should we do this same check before adding to `webmentions.json`? That would make the curation easier when we fetch new ones.

jgerigmeyer (commented):
Looks like I good start! I agree that we should curate these before publishing (and I left a comment about one way to make that a bit simpler).

On the design, my only note is that these take up a lot of space, in general, when they're not really the most important thing on the page. Could we consider hiding these by default (in an expand/collapse), or at least making the spacing and/or font sizes a bit smaller? They seem overly prominent to me.

src/filters/mentions.cjs
jgerigmeyer: Adjusted in https://github.com/oddbird/oddleventy/commit/31d21fca24310af09f2b0c6fcff30de9b578204e

src/filters/mentions.cjs
jgerigmeyer: Implemented in https://github.com/oddbird/oddleventy/commit/316cbd40f2ee5d0bedba8850f9206554ebfbdd11

jgerigmeyer:
I did an initial pass at curating them in https://github.com/oddbird/oddleventy/pull/253

mirisuzanne:
Everything here is looking good to me. I think the main open question is for @SondraE from @jgerigmeyer:

> On the design, my only note is that these take up a lot of space, in general, when they're not really the most important thing on the page. Could we consider hiding these by default (in an expand/collapse), or at least making the spacing and/or font sizes a bit smaller? They seem overly prominent to me.

It wouldn't be hard to put mentions inside a disclosure widget - just a matter of determining how that should look. I don't think we have anything similar on the site yet, where an entire content section can expand/collapse.

SondraE:
@mirisuzanne @jgerigmeyer 
> It wouldn't be hard to put mentions inside a disclosure widget - just a matter of determining how that should look. I don't think we have anything similar on the site yet, where an entire content section can expand/collapse.

In my original mockup proposal, I included a Load More Mentions button. Would you all prefer an expand/collapse instead? Is that easier to build?
Screen Shot 2022-10-07 at 1 52 19 PM

I agree that the webmentions section seems too prominent. No matter the solution for showing mentions, could we move to H4 for Name, xsmall for date, and small for mention text?

mirisuzanne:
> In my original mockup proposal, I included a Load More Mentions button. Would you all prefer an expand/collapse instead? Is that easier to build?

Expand/collapse is simpler to build, because it doesn't rely on JS - there's built in HTML functionality for it. Load more isn't super difficult. The difference that stands out to me is: I'm used to 'load more' meaning 'we have a lot of this thing and can't show them all' while disclosure means something more like 'you may not care about this extra content'. So it feels like a surprising UI to use here. There's no reason to think the first comments are the best comments, or more relevant. 

> I agree that the webmentions section seems too prominent. No matter the solution for showing mentions, could we move to H4 for Name, xsmall for date, and small for mention text?

That's not hard. We wouldn't actually use h4 - heading levels are meaningful, and we shouldn't skip numbers - but we can adjust the font sizes however we want.

jgerigmeyer:
@mirisuzanne @SondraE I'd propose we start with smaller fonts and see what that looks like, then consider expand/collapse (potentially in a new PR if small font seems publishable).

stacyk:
@jgerigmeyer @SondraE @mirisuzanne I was going to re-review and I saw there were a few style items left here so I edited the sizes according to the conversation and also made the space between a little smaller. 
Screen Shot 2022-10-11 at 2 48 59 PM

stacyk (approved):

mirisuzanne (commented):
@stacyk thanks for doing that - looks good to me!

PR #265, opened 2022-10-25
Remove Yarn PnP

jgerigmeyer:


jgerigmeyer:
[Remove zero installs](https://trello.com/c/k8dLysDo/197-remove-zero-installs)

netlify.toml
```diff
@@ -1,23 +1,17 @@
...
-    # - https://github.com/netlify/build/issues/1535
-    package = "/netlify-plugins/netlify-plugin-11ty"
+    package = "netlify-plugin-11ty"
 
     [plugins.inputs]
         cache_img = "./assets/images"
         cache_assets = false
 
+[[plugins]]
+    package = "@netlify/plugin-a11y"
```
jgerigmeyer, line 13: @mirisuzanne @oluoluoxenfree This allows us to now configure the plugin, if we decide we don't want the default settings: https://github.com/netlify-labs/netlify-plugin-a11y#configuration

mirisuzanne: Great! Since we always review changes in a PR with a deploy preview, I think we should turn off `failWithIssues`. It's useful for us to have deploy previews that build, even when there are still changes needed.

mirisuzanne (approved):

package.json
```diff
@@ -57,14 +57,14 @@
     "@11ty/eleventy-plugin-syntaxhighlight": "^4.1.0",
     "@babel/core": "^7.19.6",
     "@babel/preset-env": "^7.19.4",
+    "@netlify/plugin-a11y": "^1.0.0-beta.1",
```
jgerigmeyer, line 60: @mirisuzanne It looks like the build consistently fails with this newer version of the plugin because Pa11y times out (and we have no way of increasing the timeout). It also seems like this plugin isn't really under active development. 😞

https://app.netlify.com/sites/oddleventy/deploys/63583f8f8d55e60008d3026f

I think our options would be:

1. Revert to the older v0.0.12 (released in May 2020)
2. Try only checking a subset of files, instead of all HTML files in our (large) site

mirisuzanne: oh hmm. Well, I think it would be fair to test a sample like the pages in our main nav - and that's giving us a decent overview of some list and detail pages? Could add to the list from there, if there are other things we want to test.

jgerigmeyer: @mirisuzanne Looks like that worked, and found a lot of issues! https://app.netlify.com/sites/oddleventy/deploys/63584bb55e143c0009419955

jgerigmeyer: Merging this; added a new card to address the warnings: https://trello.com/c/zGdKgfhj/198-address-a11y-warnings

PR #268, opened 2022-10-28
Use Netlify redirects to host docs

jerivas:
Todo:

- [x] release Herman
- [x] release Accoutrement
- [x] release True
- [x] release Blend
- [x] release CCS
- [x] release Susy

I got this idea form the [Netlify forum](https://answers.netlify.com/t/support-guide-can-i-deploy-multiple-repositories-in-a-single-site/179):

1. Each project gets a new shadow repo to host the built docs, instead of including them in the main repo or the npm package
2. Each of these shadow repos also get a Netlify site that hosts the docs
3. Oddleventy uses [Netlify redirects](https://docs.netlify.com/routing/redirects/rewrites-proxies/#proxy-to-another-netlify-site) to render the per-project docs in specific folders under the original oddbird.net domain.

Right now I have created and populated the[ herman-docs repo](https://github.com/oddbird/herman-docs) manually, so the documentation for Herman should still work in this deploy preview. Up next we need to configure each project repo to push the built docs to the shadow repo, then we can merge this PR.


jerivas:
[Automate OSS project documentation from GH instead of npm.](https://trello.com/c/3ln0ndeQ/77-automate-oss-project-documentation-from-gh-instead-of-npm)

jerivas:
@jgerigmeyer looks like the redirect/rewrite is working. If you go to https://deploy-preview-268--oddleventy.netlify.app/herman/docs/ and inspect the source, you'll see the comment that I added [here](https://github.com/oddbird/herman-docs/commit/0f30fb37c6b20c7515090cb77fe98efd6cd425d8), so it's truly one site being served from another.

I'll work on automating the docs generation from Herman releases. Once we have a good setup for Herman we can replicate it on the other repos and merge this PR.

Let me know what you think of this solution.

jgerigmeyer (commented):
@jerivas This approach looks good to me! Two thoughts:

1. Would there be a way to get these redirects working in local development too? Right now it seems to result in a 404.
2. For generating the docs, I think ideally we'd want them rebuilt every time there is a "latest" release created on GH.

jerivas:
> Would there be a way to get these redirects working in local development too? Right now it seems to result in a 404.

I'll look into proxying the docs from their Netlify sites using eleventy's server. Or did you mean something local?

>For generating the docs, I think ideally we'd want them rebuilt every time there is a "latest" release created on GH.

I'm experimenting with [release triggers](https://github.com/jerivas/upgrades/blob/main/.github/workflows/publish-docs.yml#L4-L5) in my sandbox repo and it's working fine so far. Should have a PR ready for Herman tomorrow.

jgerigmeyer:
> > Would there be a way to get these redirects working in local development too? Right now it seems to result in a 404.
> 
> I'll look into proxying the docs from their Netlify sites using eleventy's server. Or did you mean something local?

Yeah, that's what I meant -- just so the links work in local development too. But that's a pretty minor point, so if it's not simple we could just replace those internal links with external links so they always work.

jgerigmeyer:
> > For generating the docs, I think ideally we'd want them rebuilt every time there is a "latest" release created on GH.
> 
> I'm experimenting with [release triggers](https://github.com/jerivas/upgrades/blob/main/.github/workflows/publish-docs.yml#L4-L5) in my sandbox repo and it's working fine so far. Should have a PR ready for Herman tomorrow.

It looks like we maybe want the `released` type instead of `published`? https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-42

jgerigmeyer (approved):
@jerivas This looks good, but I think we wait to merge this until all the relevant repo PRs are merged first, yes?

jgerigmeyer (approved):

PR #269, opened 2022-10-29
Cache images in local development.

jgerigmeyer:
11ty [`--incremental` builds](https://www.11ty.dev/docs/usage/incremental/) are still fragile (don't seem to update properly when JS/CSS are changed), and don't speed things up much anyway. The real slowness culprit is our image processor (https://www.11ty.dev/docs/plugins/image/). I've played around with various caching options, but it turns out the actual image processing itself isn't slowing things down much (and we're already using the available caching options). Simply the overall number of images we have on the site (more than 500?) means that generating the HTML for each image (even with data populated from the cache) takes 30-60 seconds.

This PR adds a dumb JSON file "cache" for images. If we already have the output requested for a given image (either the image path, or the full image markup -- depending what argument was passed to the filter), we return that and bypass the image plugin entirely. If we don't have the output for that image, then we process the image normally.

I decided to git-ignore the cache file, to ensure that the images are built locally initially for every user. When a new image is added, it should be processed and added to the local cache. *But if an existing image is replaced with a different image with the same path/name, this won't be updated locally unless the cache file (`src/filters/image_cache.json`) is deleted, or the `yarn build:images` command is run to re-process images and re-generate the cache file.*

On Netlify, the cache is ignore entirely, and all images are processed.

Thoughts, or other suggestions?


jgerigmeyer:
[As an impatient person, I'd like OddSite to reload after making changes much faster.](https://trello.com/c/VpcjUJ5U/187-as-an-impatient-person-id-like-oddsite-to-reload-after-making-changes-much-faster)

jgerigmeyer (commented):
This is ready for code and functional review 🏁

jerivas (approved):
A very smart solution! In my case the time from saving a Sass file to reloading the browser went from ~40s to ~10s. I'm sure we can do more to bring that down, but a 4x speed up is already an excellent improvement 🚀

stacyk (approved):
This is so much faster @jgerigmeyer! Thank you!

sanajaved7:
Wow, this is such a huge improvement! ✨🎉  With the recent footer changes, the reload usually took about 35-45 seconds which adds up when you're testing different things so the fact that its down to ~5 seconds for me is awesome!

sanajaved7 (approved):

mirisuzanne (approved):
I'll download and try it out, but the idea/code look good to me.

mirisuzanne:
oh wow, that's like a 3s rebuild time for me on the index page. That's a huge improvement. I need to port this change to my site.

PR #273, opened 2022-11-02
Fix a few more broken links.

jgerigmeyer:


jgerigmeyer:
Fixed a few more broken links, and added a different tool that can be used (locally or via GH action) to check links (the old tool just timed out every time).

For now I didn't set this up to run automatically, but we could do that if we want.

jerivas (approved):
Good to know these tools are available. I wonder if an automated run every 30 - 180 days could be useful

PR #305, opened 2023-01-09
remove honeypot field

oluoluoxenfree:
Removing the honeypot field for a11y reasons; will look and see if there's another way to do a similar thing without using aria.


oluoluoxenfree:
[Address a11y warnings.](https://trello.com/c/zGdKgfhj/198-address-a11y-warnings)

content/_includes/contact.macros.njk
```diff
@@ -62,11 +62,6 @@ params:
     required=true
   ) }}
 
-  <label data-honey aria-hidden="true">
-    <span>Don't fill out this field:</span>
-    <input name="subject" />
-  </label>
```
jgerigmeyer, line 68: If we want to remove this, we can also remove the `netlify-honeypot` attr from the `<form>` above.

jgerigmeyer (approved):
One minor comment, but I'm okay with this -- I guess we'll see if we start getting more spam submissions?

mirisuzanne (approved):
This seems fine to try. I agree, we can judge how much spam increases. 

We may also want to open an issue on the plugin, tho. I don't think this should throw an error?

oluoluoxenfree:
@mirisuzanne @jgerigmeyer Do you reckon this would be an issue with pa11y/the upstream provider rather than the netlify plugin itself? Happy to raise an issue.

mirisuzanne:
@oluoluoxenfree good point - yeah, it seems likely an issue with pa11y? Worth at least raising the question with them.

oluoluoxenfree:
Rereading this I'm not sure they'd consider it an issue; the field has to be actually disabled for it to pass: https://dequeuniversity.com/rules/axe/4.2/aria-hidden-focus?application=axeAPI

mirisuzanne:
@oluoluoxenfree oh, ok, interesting. That makes some sense.

PR #306, opened 2023-01-13
[sass-config-two] Split scale and animation into separate modules

dvdherron:
## Steps to test/reproduce
- make sure animations and sizing variables still behave as expected. 

## Show me


src/scss/config/animation/_index.scss
```diff
@@ -0,0 +1,42 @@
+@use '~sassdoc-theme-herman/scss/utilities' as herman;
+@use '../tools';
+@use 'changes';
+@use 'times';
+@use 'sass:meta';
+@forward 'changes';
+@forward 'times';
+
+/// # Animation Config
```
stacyk, line 9: We don't need `@include herman.add ...` unless we are wanting Herman to show a visual preview, right?

src/scss/config/scale/_spacing-sizes.scss
```diff
@@ -0,0 +1,11 @@
+$gutter: 0.75rem;
+$newline--default: $gutter;
+$newline: $newline--default;
+$gutter-plus: calc(var(--gutter) * 1.5);
```
stacyk, line 4: I'm curious... We can reference a Sass variable like line 2 here or we can use the CSS variable like line 4. If we do the former, we get this type of visual preview which gets grouped together:
Screenshot 2023-08-08 at 11 13 22 AM

vs if we use the css custom properties we see this:
Screenshot 2023-08-08 at 11 14 05 AM

I know we are using the Sass variables on OddBooks. If we use the CSS custom properties, it will show the values as defined so for double-gutter it will show calc(var(--gutter) * 2) as opposed to the Sass variables which will calculate and show the result which is 1.5rem.

I don't know if there are other reasons or advantages (I know we discussed this a while back and this project was using a mix of both custom properties and Sass Variables so I thought I'd check to see if there was any opinions. I do like seeing both the math (showing the custom props) and the calculated value, which you get if we use the Sass Variables. I don't love referencing Sass variables from other files because of the namespacing, but that isn't a big deal. The maintainer in me like that the namespace tells you where it is coming from, but author in me doesn't love the few extra things to check/write ¯\_(ツ)_/¯

src/scss/initial/_root.scss
```diff
@@ -3,6 +3,9 @@
 
 html {
   @include config.font-family('body');
+  @include config.ratios--;
+  @include config.times--;
+  @include config.changes--;
```
stacyk, line 8: I didn't think that ratios, times and changes would belong in their own initial/* partials like the scale and colors partials but I am used to seeing all these included at once.

stacyk (commented):
@mirisuzanne this is ready for you to review. [Here is the Trello card](https://trello.com/c/M6Vy6GtZ/184-sass-config-continued-split-scale-and-or-animation-into-separate-modules ).

src/scss/config/_layout.scss
```diff
@@ -28,3 +29,14 @@ $z-index: ('bump', 'dropdown', 'skip-link');
     transform: translateY(calc(var(--offset) * -1));
   }
 }
+
+// Breakpoints
+@each $name, $size in meta.module-variables('page-sizes') {
```
mirisuzanne, line 34: Looking at the whole list of page-sizes below, I'm not sure it makes sense to turn them all into breakpoints? Some of them are things like logo dimensions.

src/scss/config/scale/_spacing-sizes.scss
mirisuzanne: I don't think we should make the decision based too much on how it looks in the style guide. I'd probably lean towards capturing relationships in custom properties. I think that's what you're suggesting as well? Especially for something like `newline` and `newline-default` - we want that relationship to exist in the CSS, so when you update the default, the other one also updates.

mirisuzanne (approved):
looks good! Left a few small comments.

src/scss/config/_layout.scss
dvdherron: @mirisuzanne I moved `logo-item` and `grid-item` to theme-sizes in https://github.com/oddbird/oddleventy/pull/306/commits/6bf56ceeaf8da19b0c67605c9341586443c3d6ae.

And there was a card for relocating the `extra` and `main-min` variables that I figured would be easier knock out here too. I don't think we are using those sizes as breakpoints either. Do you think they make sense in `theme-sizes` as well? see https://github.com/oddbird/oddleventy/pull/306/commits/2410ea816c4602d741265868ccf969fb96a9c964

src/scss/config/scale/_spacing-sizes.scss
dvdherron: I changed `newline` and `newline-default` back to custom properties to capture the sizing relationships as suggested in https://github.com/oddbird/oddleventy/pull/306/commits/6bf56ceeaf8da19b0c67605c9341586443c3d6ae

src/scss/config/scale/_theme-sizes.scss
```diff
@@ -1,4 +1,11 @@
...
+@use 'page-sizes';
+
 $page-margin: calc(var(--quarter-shim) + 4vw);
 $border-width: medium;
 $radius: var(--half-shim);
 $grid-hero: calc(var(--gutter) * 8);
+$logo-item: 10rem;
+$grid-item: 16rem;
+$main-min: calc(100vw - var(--page-margin) * 2);
+$extra: calc(var(--wide-page) - var(--page) / 2);
```
stacyk, line 11:
```suggestion
$extra: calc((var(--wide-page) - var(--page)) / 2);
```

Otherwise the division happens first

src/scss/config/scale/_theme-sizes.scss
```diff
@@ -1,4 +1,11 @@
+@use 'sass:math';
+@use 'page-sizes';
```
stacyk, line 2: these aren't necessary with what is currently in the file. 
```suggestion
```

stacyk (commented):
@dvdherron just a tiny bit more cleanup then this is good to go!

src/scss/config/scale/_theme-sizes.scss
dvdherron: good catch. fixed in https://github.com/oddbird/oddleventy/pull/306/commits/088d051b4dd722f80ad7747508b815562102895c

src/scss/config/scale/_theme-sizes.scss
dvdherron: whoops. removed in https://github.com/oddbird/oddleventy/pull/306/commits/088d051b4dd722f80ad7747508b815562102895c

dvdherron:
[Relocate sizing variables](https://trello.com/c/SAdnDclK/155-relocate-sizing-variables)

stacyk (approved):
LGTM!

stacyk:
@mirisuzanne did you want to take another look at this before we merge?

mirisuzanne (approved):

PR #310, opened 2023-01-18
aria change removing page label

oluoluoxenfree:
## Steps to test/reproduce

## Show me


oluoluoxenfree:
[Address a11y warnings.](https://trello.com/c/zGdKgfhj/198-address-a11y-warnings)

jgerigmeyer (approved):
I notice we have other `aria-label` attrs in our site ("all the birds", "site", "OddBird.Net", "contact", and "primary") -- are those not issues?

mirisuzanne (approved):

netlify.toml
```diff
@@ -23,6 +23,8 @@
           "/about/index.html",
           "/contact/index.html",
         ]
+        ignoreElements = '#search'
+        ignoreGuidelines = ['color-contrast']
```
jgerigmeyer, line 27: @oluoluoxenfree Instead of ditching the plugin entirely, what do you think about this solution -- selectively ignoring the errors we don't want to address? Seems like we could even switch to `failWithIssues = true` so we get notified about new issues?

oluoluoxenfree:
> I notice we have other `aria-label` attrs in our site ("all the birds", "site", "OddBird.Net", "contact", and "primary") -- are those not issues?

AFAIK it's only an issue when it's too general, so "Site" might be but maybe it couldn't automatically pick it up?

mirisuzanne:
> AFAIK it's only an issue when it's too general, so "Site" might be but maybe it couldn't automatically pick it up?

@oluoluoxenfree Want to look through those and propose the best option in each case? eg 'keep it as is', 'keep it but add detail', or 'remove the aria attribute'?

netlify.toml
oluoluoxenfree: @jgerigmeyer should i add this to this PR?

jgerigmeyer: @oluoluoxenfree Sure! It won't pass until we merge #305 in as well, but that's okay.

oluoluoxenfree: Do people prefer `failWithIssues`?

oluoluoxenfree:
Reading [this](https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html) and [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) it should only be only interactive elements that need some sort of instructions for use but can't have visible ones (e.g. a close button that's an x); I will check but I don't think it applies to our current usage.

oluoluoxenfree:
| aria-label | keep/discard/change | why? |
|---|---|---|
|  'all the birds' | 🚮  |  not interactive; exchange for image alt if we think it's useful as a non-decorative element  |  
| 'site' | 🚮  |  not interactive, not providing extra info |  
|  'oddbird.net' | ♻️ | I think the intention of this one is to label the link takes you back to the home page? In that case can edit and keep |
| 'contact' | ♻️/✅ | change to 'contact form' maybe, otherwise fine as is, as does have an interactive element |
| 'primary' | ♻️ | is this to show there's more than one menu? could be 'primary menu' or 'menu 1' or 'main menu' |

mirisuzanne:
- 'all the birds': I assume we used it there because svg elements don't have an alt attribute. And we're building that svg dynamically in the page, so we can't switch that over to an image element. If we want alt text there (maybe we don't?), we'll need to find a different solution for it.
- 'primary': yeah, `nav` regions show up in landmark navigation (so do headers, which is maybe why we added `site`?). So the idea with these was probably to give different navigation elements distinct names. And as I understand, 'navigation' is already added to the end of a region when announced, which is why we don't say that explicitly in the label.

Using a firefox plugin to show the landmarks on a page, this is what I see on our live site:

Screenshot 2023-02-15 at 9 48 46 AM

oluoluoxenfree:
@mirisuzanne oh wow i didn't know that about svgs! I wonder what the history of that is; maybe they were intended to be decorative at first? 🤔 I assume that means we don't need a label for that at all.

I'm not sure what the complementary is about in the site landmarks, having a look into that.

oluoluoxenfree:
The top complementary (an aside is what that means, TIL) is the black lives matter banner which i assume need a label of some sort.

mirisuzanne:
I think the original idea with svg was that authors should use the `<title>` element, and that seems to also be part of the recommendation from this Smashing article: https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/#basic-alternative-descriptions-using-the-svg-tag-group-2. 

I think we were trying to follow the pattern suggested by MDN: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Img_role#svg_and_roleimg. I don't see that option covered in the Smashing article, or know which one is better.

I think you're right we could give the top BLM aside a more specific region name. Maybe using aria-describedby, like we do with the main area?

oluoluoxenfree:
I only just saw your comment, will move to described by, i was just unsure if we need a new element to tag in if we do that, e.g.

```
<... aria-describedby="tag1">
  <id="tag1">???<>
```

content/_includes/site/alert.njk
```diff
@@ -1,7 +1,7 @@
 {% set alert = collections.all | findPage('data.info_slug', 'alert') %}
 
 {% if alert and (alert.data.live) %}
-  <aside data-layout="alert" data-ccs-colors="dark">
+  <aside data-layout="alert" data-ccs-colors="dark" role="banner" aria-label={{ alert.data.aria }}>
```
mirisuzanne, line 4: Why `role=banner`? I think the Complementary role that comes from the aside is probably fine here?

content/_meta/info/alert.md
```diff
@@ -2,6 +2,7 @@
 live: true
 info_slug: alert
 title: alert
+aria: BLM
```
mirisuzanne, line 5: This approach makes sense. Maybe best to spell it out rather than abbreviating?

oluoluoxenfree: I need to work out how to stop it from only taking the first word in that case, will do :)

content/_includes/site/alert.njk
oluoluoxenfree: Ah okay, that makes sense.

mirisuzanne:
looks good to me

mirisuzanne (approved):

PR #327, opened 2023-02-28
Fix featured posts, and use typogr consistently.

jgerigmeyer:

### DIFF
--- a/content/_layouts/base.njk
+++ b/content/_layouts/base.njk
@@ -73,10 +73,12 @@
       Skip to main
     </a>
 
+    {% typogr %}
     {% include "site/alert.njk" %}
     {% include "site/banner.njk" %}
     {% include 'page/base.njk' %}
     {% include 'site/footer.njk' %}
+    {% endtypogr %}
 
     <script type="text/javascript" src="/assets/js/scripts.js"></script>
 
--- a/content/_includes/page/base.njk
+++ b/content/_includes/page/base.njk
@@ -24,7 +24,7 @@
   {% if content -%}
     {%- set hcontent = 'e-content' if is_entry else ('p-note' if is_bird else none) -%}
     <div data-main data-typeset {{ utility.attr_if('class', hcontent)}}>
-      {{ content | typogr | safe }}
+      {{ content | safe }}
     </div>
   {%- endif %}
 
--- a/content/_includes/post.macros.njk
+++ b/content/_includes/post.macros.njk
@@ -546,7 +546,7 @@ params:
   {% if adr %}
     <i>in</i>
     <span class="adr">
-      {{ adr | typogr | safe }}
+      {{ adr }}
     </span>
   {% endif %}
 {% endmacro %}
@@ -604,7 +604,7 @@ params:
           {% for item in items | sort(attribute='tag') %}
             <li {% if pill %}class="pill-item p-category"{% endif %}>
               {{- utility.link_if(
-                content=item.tag | typogr,
+                content=item.tag,
                 url=item.url,
                 class='pill' if pill else none
               ) -}}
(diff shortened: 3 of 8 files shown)


content/_layouts/base.njk
```diff
@@ -73,10 +73,12 @@
       Skip to main
     </a>
 
+    {% typogr %}
     {% include "site/alert.njk" %}
     {% include "site/banner.njk" %}
     {% include 'page/base.njk' %}
     {% include 'site/footer.njk' %}
+    {% endtypogr %}
```
jgerigmeyer, line 81: @mirisuzanne I noticed we had double-typogr'd code all over the place (e.g. `This <span class="caps"><span class="caps">CSS</span></span> Thing`). The main change here is to remove `typogr` from where it was used manually and where it was auto-used any time we used `md` or `mdInline`, and instead wrap the entire site contents in this `{% typogr %}` shortcode. That's what we're doing on Sass Site as well.

I think the functional downside / difference is that before, the `mdInline` filter was _not_ adding `widont` in cases of 5 words or less... How important is that?

jgerigmeyer (commented):
@mirisuzanne @oluoluoxenfree I submitted this in its own PR for easier review, especially the generated diff: https://github.com/oddbird/oddleventy-built/compare/positioning-update..typogr

oluoluoxenfree:
I'm a bit confused by the one test failure as that looks very different/like it's adding a lot, otherwise looks good to me.

jgerigmeyer:
> I'm a bit confused by the one test failure as that looks very different/like it's adding a lot, otherwise looks good to me.

Fixed in https://github.com/oddbird/oddleventy/pull/327/commits/96f00379cbe2fc8a3847f7e0127202d05e17b9c5

content/_layouts/base.njk
jgerigmeyer: @mirisuzanne Nudge on this...

mirisuzanne: This approach makes sense. We'll want to check the output. The problem with applying widont on short text is that you can end up with very strange line balancing. We can see if that problem shows up in a deploy preview? I'd be tempted to remove typogr entirely, and switch to `text-wrap: balance` once it's available anywhere. https://cloudfour.com/thinks/responsive-headlines-are-about-to-get-awesome/

mirisuzanne (approved):
this looks fine, without checking the full site for any possible headline wrapping issues.

content/_layouts/base.njk
jgerigmeyer: Yeah, interesting. You can see the output diff here: https://github.com/oddbird/oddleventy-built/compare/positioning-update..typogr

mirisuzanne: Can't tell from that how it actually ends up looking, but yeah - some of those are a bit worrying. There's really no point putting widont in a 2 & 3 word headlines. It'll just lead to overflow, and doesn't really prevent anything.

jgerigmeyer: Hm, it shouldn't add anything when there are only 2 words, but it will add to 3-word headings. I don't see a great alternative at this point?

mirisuzanne: I think the alternative is we apply it less often, rather than more often. For example: only on headlines, and only when they have more than 5 words. Widows are a bit annoying, but I think they might be better than making 3-word headlines wrap after the first word?

mirisuzanne: That might be a question for @SondraE or @stacyk though

jgerigmeyer: Or we just turn off the `widont` feature?

jgerigmeyer: Disabled in https://github.com/oddbird/oddleventy/commit/1dd8a9ce8d580eda5a9bf095b39daa5940dc37f6

mirisuzanne (approved):

content/_layouts/base.njk
stacyk: I'm good with removing widont all together. IMO, it gets in the way more than it helps.

PR #357, opened 2023-05-17
[img-helpers] Add/Edit Image display utilities

stacyk:
## Description
Add/edit image helpers:

- [x] Fix img-shadow (and convert all ing-border to use ing shadow instead)
- [x] Add ability to display an image to the left of the text on large enough screens (media-pull)
- [x] Remove extend-small (it was the same as what the figure width is)
- [x] Add caption padding for images that use extend-full
- [x] Add examples to the Sample Pages / Placing Images page
- [x] Remove unused sizes and extend-left extend-right classes
- [x] Update some of the older embed.img macros to the embed.figure macro (especially the ones that were wrapped in a `<figure>` tag

## Steps to test/reproduce
Review a few pages that were updated:
- work pages ( /work/coachhub/, /work/adobe, /work/quarqnet )
- blog posts ( 2018/01/03/introducing-herman/ )
- review the examples on the Sample Placing images page: /sample/images/

## Show me

sample-images


stacyk:
[Remove or fix unused CSS classes](https://trello.com/c/f61Si0tU/214-remove-or-fix-unused-css-classes)

stacyk:
@mirisuzanne this is ready for an initial review. I kept these "helpers" as classes instead of adding a data attribute incase we wanted to apply it to something other than the embed.figure or embed.img macros. Like the contact form uses extend-small.

I also took over what was the sample page called Responsive Images to show how these examples look when applied (I think it will be helpful to have a visual guide to know what is available. 

This is not something in this PR, but it is related to this images sample page. I don't actually know what the difference is between the first 2 (one says no width and the other says 4410) but the secret sizes are the same and so is the visual display. The last one is using the "sizes" attribute but you can't tell from the page itself. It does have a different srcset size value. I feel like we should add some explanation to these images and why they are there, but I am not actually sure why they are all there.

content/sample/images.md
```diff
@@ -24,3 +24,127 @@ sub: Using nunjucks
...
+
+</div>
+
+## Extending images outside the content column
+
+To extend an image further outside of the content column's width, add one of
+the following classes: `extend-large`, or `extend-full`. This is only available
+with the `embed.figure` macro as the `embed.img` macro's class doesn't have
+access to the outer container. Also, the embed.figure macro is wider than the
+content column by default, as opposed to the `embed.img` macro.
```
mirisuzanne, line 98:
```suggestion
access to the outer container. Note that the `embed.figure` macro extends wider than the
content column by default, as opposed to the `embed.img` macro.
```

content/sample/images.md
```diff
@@ -24,3 +24,127 @@ sub: Using nunjucks
...
+
+### `embed.figure` without any entend class for comparison
```
mirisuzanne, line 100:
```suggestion
### `embed.figure` without any extend class for comparison
```

Seems like this could also be called `extend-small` - and available to other media?

content/sample/images.md
```diff
@@ -24,3 +24,127 @@ sub: Using nunjucks
...
+  class='extend-full',
+  caption='I am a caption'
+
+) }}
+
+## Add a shadow to make an image stand out from the background
+
+You can add the class `img-shadow` to the `embed.figure` and `embed.img` to
+display a shadow around the image. This doesn't work on figure if you have a
+caption as the shadow will be around that entire container.
```
mirisuzanne, line 134: Note that the shadows are not visible in dark mode?

src/scss/patterns/_alignment.scss
```diff
@@ -1,21 +1,29 @@
+@use '../config';
+
 .contain {
   display: flow-root;
 }
 
 [class*='align-'] {
-  margin: var(--shim) var(--inline-margin, 0);
+  margin: var(--shim) var(--inline-margin-right, var(--inline-margin, 0)) var(--shim) var(--inline-margin-left, var(--inline-margin, 0));
```
mirisuzanne, line 8:
```suggestion
  margin-block: var(--shim); 
  margin-inline-start: var(--inline-margin-start, var(--inline-margin, 0));
  margin-inline-end: var(--inline-margin-end, var(--inline-margin, 0));
```
A bit easier to follow? And not mixing logical/physical names?

src/scss/patterns/_gallery.scss
```diff
@@ -47,6 +47,10 @@ figcaption,
   font-size: var(--small);
   font-style: italic;
   margin-top: var(--shim);
+
+  .extend-full & {
+    padding-inline: var(--gutter-plus);
```
mirisuzanne, line 52: Should we use `var(--page-margin)` here?

src/scss/patterns/_media.scss
```diff
@@ -116,3 +116,27 @@
...
+.media-pull {
+  --column: page-start;
+
+  @include config.above('xlarge-page') {
+    --column: full-start / page-start;
+
+    grid-row: var(--rowspan);
+    justify-self: end;
+    max-width: 25vw;
+    padding-inline: calc(var(--gutter) + 2vw);
```
mirisuzanne, line 133: These numbers feel a bit magical and unclear to me. Why 25vw? Right now, this doesn't align with anything else on the page - should it? Is the extra gutter space meant to avoid our hanging link icons on headings? If you use this next to a paragraph rather than a heading, does the extra space still make sense?

src/scss/patterns/_sizes.scss
```diff
@@ -1,28 +1,7 @@
...
-}
-
-.extend-right {
-  --column: page / wide;
-
-  width: 100%;
-}
-
 // Extend thing past the width of its container on both sides
 .extend-small {
```
mirisuzanne, line 6: This one isn't documented?

mirisuzanne (commented):
Good stuff here! I just have a few questions about some of the details.

content/sample/images.md
```diff
@@ -24,3 +24,127 @@ sub: Using nunjucks
...
+    img: 'blog/2023/graph.jpg',
+    alt: 'sketched graph on paper'
+  }],
+  class='media-pull',
+  attrs={'style': '--rowspan: span 2'}
+) }}
+
+## `media-pull` will position an image next to the text on wide enough screens
+
+The `embed.figure` macro is the only one that has access to the outer element in
```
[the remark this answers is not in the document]
stacyk: yes. the embed.image is always wrapped in a <p> though... so you don't have access to add the class to the <p> that it is wrapped in unfortunately. I should have clarified that this is available on embed.figure but not embed.img (I didn't test it works on other macros, only those two).

stacyk (commented):
@mirisuzanne I think I covered all/most of your comments minus the one I had a question about (extend-small)

content/sample/images.md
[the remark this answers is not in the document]
stacyk: While I agree that the image in a `<p>` tag isn't ideal, I thought we were stuck with that for reasons.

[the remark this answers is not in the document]
mirisuzanne: Well, we're maybe stuck with that in markdown specifically, unless we write our own transform (we probably could, but it's not super simple). But not all of our images are used in markdown - we sometimes use these macros in nunjucks files. So it seems to me like we should document the actual limitation here, and think of extending images as 'valid, but dangerous inside markdown until we find a fix'. 

Here's a fix that might work in modern browsers:

```css
p:has(> picture:only-child) { display: contents; }
```

src/scss/patterns/_sizes.scss
mirisuzanne: See my case above that it may still be useful in other situations, like in nunjucks. If it's a thing we do sometimes, I think it makes sense to have a standard way of doing it.

mirisuzanne:
> This is not something in this PR, but it is related to this images sample page. I don't actually know what the difference is between the first 2 (one says no width and the other says 4410) but the secret sizes are the same and so is the visual display. The last one is using the "sizes" attribute but you can't tell from the page itself. It does have a different srcset size value. I feel like we should add some explanation to these images and why they are there, but I am not actually sure why they are all there.

Yeah, these were for testing the responsive image stuff as I was writing it - but they're probably outdated, and were never well documented. We should add a story to actually document our responsive image handling, but these don't seem useful, so I think they can just be removed.

content/sample/images.md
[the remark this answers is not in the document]
stacyk: <3 `:has`  

When will Firefox support this? I can't wait to use it more.

mirisuzanne (approved):
looks great. I would still probably vote for documenting how to extend images, and how to extend-small, while also noting that it might not work in markdown. If the `has()` trick even works for some browsers - it's not bad for that to be a progressive enhancement.

stacyk:
@mirisuzanne I updated the sample images page to include the updated guidelines for using the img/figure macros and added that little `p:has(> picture:only-child) { display: contents; }` snippet along with a few styles for fallbacks. You can see it not working in Firefox and working in the others. 

Firefox:
Screenshot 2023-06-13 at 5 36 45 PM

Chrome: 
Screenshot 2023-06-13 at 5 38 08 PM

I removed the srcset examples and added a note to this existing image handling card: https://trello.com/c/p8iLneXl/117-document-responsive-image-handling-and-any-other-image-guidelines-that-arent-yet-on-the-sample-pages

content/sample/images.md
```diff
@@ -1,26 +1,173 @@
...
+
+To center align an image add the class `align-center`. This is only applicable
+if your image is smaller than the column width.
+
+</div>
+
+## Extending images outside the content column
+
+To extend an image further outside of the content column's width, add one of
+the following classes: `extend-small`, `extend-large`, or `extend-full`.
```
stacyk, line 92: @mirisuzanne I added `extend-small` back in this documentation since we can add this to the embed.img macro now. Also it is useful for showing how extend-small's width is the same as the embed.figure macro by default so it only makes sense to use it on embed.img.

content/sample/images.md
```diff
@@ -1,26 +1,173 @@
...
+
+When using the `embed.img` macro in a markdown file, the image is placed
+within a `<p>` tag. We have a CSS helper that targets this nested image with
+`display: contents` in order to add these layout classes in [most modern
+browsers](https://caniuse.com/css-has).
```
stacyk, line 97: I added this issue to the trello story https://trello.com/c/PkgEjxR2/122-when-you-use-the-embedfigure-macro-it-adds-an-empty-p-p-before-and-after

content/sample/images.md
```diff
@@ -1,26 +1,173 @@
...
 
-### embed (width: 4410):
+### `media-pull` with the `embed.img` macro
+
+When using the `embed.img` macro in a markdown file, the image is placed
+within a `<p>` tag. We have a CSS helper that targets this nested image with
+`display: contents` in order to add layout classes which will work in [most
+modern browsers](https://caniuse.com/css-has). When unsupported, the image will
+appear centered within the content column.
+
```
stacyk, line 41: @mirisuzanne I added this second example specifically for testing the embed.img macro since it doesn't work in Firefox yet.

src/scss/patterns/_media.scss
```diff
@@ -129,7 +129,20 @@
...
+    // include the margin when used with embed.figure macro
+    &.embed {
+      --media-pull-margin: var(--page-margin);
+    }
+
+    // include the margin when used with embed.img macro on supported browser
+    // fallback will show an image that takes up the full content column
+    @supports (selector(:has(*))) {
+      --media-pull-margin: var(--page-margin);
+    }
```
mirisuzanne, line 145: good thinking!

mirisuzanne (approved):
looks great!

PR #392, opened 2023-08-01
Feature winging it post on homepage

jgerigmeyer:

### DIFF
--- a/content/_includes/post.macros.njk
+++ b/content/_includes/post.macros.njk
@@ -768,12 +768,8 @@ note: |
   posts from a collection for the
   homepage "featured posts" section
 params:
-  type:
-    type: string
-    note: one of several post "types" to display (types are defined in the `taxonomy.yaml` data file)
   collections:
     type: 11ty collections object
-    default: '[]'
   limit:
     type: number
     default: 4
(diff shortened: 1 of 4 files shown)


jgerigmeyer:
[Feature Winging It posts on the home page](https://trello.com/c/MW1Sg9Vd/223-feature-winging-it-posts-on-the-home-page)

mirisuzanne (approved):
Code change looks great. I don't know how we decide which to article to stop featuring - if we don't just want them to fall off the list by date. I'll leave that question to @SondraE

stacyk (approved):

SondraE (approved):
Regarding Miriam's question about how to decide which article to stop featuring:

I think it would be fine to have the most recent items on the homepage. Perhaps we go by date until a situation arises where we want something different.

jgerigmeyer:
> I think it would be fine to have the most recent items on the homepage. Perhaps we go by date until a situation arises where we want something different.

@mirisuzanne @SondraE So you're saying we should just always feature the 4 most recent posts instead of using the manual `home` setting? Or that if there are more than 4 `home` posts, we should show the 4 most recent ones of those?

mirisuzanne:
I was suggesting the latter, but would be happy with the former as well. Or making it an explicit `home:false` option, if we want to exclude some things?

content/blog/2021/containerqueries.md
```diff
@@ -17,7 +17,6 @@ summary: |
   and proposing for years, has finally made its debut in a browser. Well, sort
   of. Here we'll explain what container queries are, how they work, and what
   other features they might come with once fully supported in browsers.
-home: true
```
jgerigmeyer, line 20: I removed all the `home` settings for now, but let me know if you want any of the 4 featured ones removed, or another post pinned.

jgerigmeyer (commented):
@mirisuzanne @SondraE Ready for another round of review.

Posts are included by default, but can be omitted with `home: false`, or pinned with `home: pin` (or `home: pinned`).

SondraE (approved):
lgtm

PR #419, opened 2023-09-29
Get latest webmentions

mirisuzanne:
## Description

I setup mastodon webmentions via Bridgy Fed, but most of the activity there has been either:

- likes and reposts, which we don't track
- not specific to an article on oddsite


content/_data/webmentions.json
```diff
@@ -1,6 +1,29 @@
...
   "children": [
+    {
+      "type": "entry",
+      "author": {
+        "type": "card",
+        "name": "Ana Rodrigues",
+        "photo": "https://webmention.io/avatar/files.mastodon.social/ecf9b50ba4c4460d358846b6c016e420ad56e1abdb2cfb6b8c1e4ae82cded723.png",
+        "url": "https://mastodon.social/@ohhelloana"
+      },
+      "url": "https://mastodon.social/@ohhelloana/110281440017986097",
```
jgerigmeyer: It's weird that it picked up this reply, but not the original post: https://mastodon.social/@matthiasott/110277352371004975

@mirisuzanne I removed this since it doesn't make much sense out of context.

jgerigmeyer (approved):

content/_data/webmentions.json
mirisuzanne: Yeah, not sure what's up with that.

PR #448, opened 2023-11-14
Add shoptalk and smashing video

mirisuzanne:

### DIFF
--- a/content/blog/2023/sqlalchemy-for-django-developers.md
+++ b/content/blog/2023/sqlalchemy-for-django-developers.md
@@ -1,4 +1,5 @@
 ---
+card: feature
 title: SQLAlchemy for Django Developers
 author: ed
 date: 2023-10-23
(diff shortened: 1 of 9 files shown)


mirisuzanne:
I'm not sure when/if to feature various things. I did go back and feature learning with Jason, because that one seemed popular.

SondraE (approved):
LGTM
As for when to feature, I think it would be good to feature things we want people to notice and also things that are popular so people coming to the site can find them more easily.

By that metic we could also feature: Winging It htmx, SQLAlchemy article, and Unintuitive Objects article.
We could un-feature [When to Choose a Responsive Web App](https://www.oddbird.net/2023/07/11/when-to-choose-a-responsive-web-app/).

mirisuzanne:
> By that metic we could also feature: Winging It htmx, SQLAlchemy article, and Unintuitive Objects article.

@SondraE The old or new SQLAlchemy article? I assume the new one?

jgerigmeyer (approved):
Looks good. If we want, I'll also note that we have the option to "pin" certain posts to the homepage, otherwise it'll just cycle through the most recent.

PR #449, opened 2023-11-16
Article: html web components are just JS

mirisuzanne:
## Description

An article about web components


jgerigmeyer (approved):

mirisuzanne:
@jgerigmeyer I think I addressed your comments, and made a few other small changes.

jerivas (approved):
Very interesting read! I like the idea of using different tools to generate markup on the server and augmenting elements in-browser

PR #454, opened 2023-11-22
make sure post thumbs are same height

stacyk:
## Description
Some image ratios will end up looking shorter in our post lists. This will fix that by giving all thumbs a fixed height.

## Steps to test/reproduce
Compare the Winging it post on the [live site homepage](https://oddbird.net) to [this deploy preview](https://deploy-preview-454--oddleventy.netlify.app/) (or see screenshots) 
Compare the [live site's 2022 "State of CSS Frameworks"](https://oddbird.net/blog/2/) post thumb in the post list to the one [in this preview](https://deploy-preview-454--oddleventy.netlify.app/blog/2/) (or see screenshots)

## Show me

Post List Before
post-list-before

Post List After
post-list-after

Before
home-before

After
home-after

### DIFF
--- a/src/scss/components/_post.scss
+++ b/src/scss/components/_post.scss
@@ -19,7 +19,10 @@
   grid-area: post;
   grid-row-gap: var(--shim);
   grid-template:
-    [hero-start] 'left type type' auto
+    [hero-start] 'left type type' var(
+      --post-thumb-height,
+      var(--post-thumb-small)
+    )
     [hero-end] 'left main right' 1fr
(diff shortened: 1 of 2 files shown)


stacyk:
[Consider giving the post grid article grid hero area a fixed height so images look consistent on post grid pages](https://trello.com/c/WdcjO5hj/234-consider-giving-the-post-grid-article-grid-hero-area-a-fixed-height-so-images-look-consistent-on-post-grid-pages)

mirisuzanne (approved):
This works for me. I suppose the other solution could be setting an aspect ratio? But I'm not sure that would be better for any reason. lgtm

Issue #482, opened 2024-01-02
Use <lite-youtube> for youtube embeds


stacyk:
See https://www.zachleat.com/web/youtube-embeds/

https://github.com/paulirish/lite-youtube-embed

The article recommends wrapping the embed in `is-land`, which would give us lazy loading, and even further decrease the initial page load. 

I think the benefits for page load are worth considering. The cost would be the additional dependencies, but neither lite-youtube-embeds or is-land have their own dependencies.

A brief comparison of https://oddbird.ing/2017/01/31/css-day/, which has 7 videos.

Current: 

![Image](https://github.com/oddbird/oddleventy/assets/167908/22301ea9-491c-4f19-b109-20575e2608cd)

With `is-land` set to `on:visible` and `lite-youtube-embed`-

![Image](https://github.com/oddbird/oddleventy/assets/167908/bd88c840-7d91-4ab0-b18b-eb93535ab359)

That's a fairly significant difference.

jgerigmeyer:
Assigning to @jamesnw to merge or clarify what's remaining.

PR #485, opened 2024-01-02
Lite Youtube

jamesnw:
## Description
Replaces iframe Youtube embeds with `lite-youtube`, wrapped in an `is-land` component. The default lite-youtube poster images are rather low quality, so this uses the 11ty open graph API to get a higher quality image. We have a few options on the poster frames-

- host our own version of the [API](https://www.11ty.dev/docs/services/opengraph/)
- rely on 11ty's instance of the API (as in this PR)
- Provide a way for users to specify a poster image that we host in the site
- Have crappy images

I'm also open to feedback on the video macro- matching on the string in the video source doesn't feel right, but it allows us not change existing embeds (although there aren't many), and gives us a consistent interface for all embed.videos. Alternatively, we could add an `embed.youtube(id)` macro.

## Related Issue(s)

#482 

## Steps to test/reproduce

Compare [2017/01/31/css-day/](https://deploy-preview-485--oddleventy.netlify.app/2017/01/31/css-day/) to https://oddbird.ing/2017/01/31/css-day/


jgerigmeyer (commented):
Looks good!

> The default lite-youtube poster images are rather low quality, so this uses the 11ty open graph API to get a higher quality image. We have a few options on the poster frames-
> 
> * host our own version of the [API](https://www.11ty.dev/docs/services/opengraph/)
> * rely on 11ty's instance of the API (as in this PR)
> * Provide a way for users to specify a poster image that we host in the site
> * Have crappy images

Hm, what's your leaning here? It seems like options 1 or 2 would be preferred -- is there a downside to using the 11ty hosted version? I guess hosting our own is probably more robust, and doesn't look like it'd be difficult.

> I'm also open to feedback on the video macro- matching on the string in the video source doesn't feel right, but it allows us not change existing embeds (although there aren't many), and gives us a consistent interface for all embed.videos. Alternatively, we could add an `embed.youtube(id)` macro.

I think string matching is okay in this case, although adding a `type: 'youtube'` arg would be fine too -- we could be more strict about matching the entire string, e.g. `https://www.youtube.com/embed/`.

jamesnw:
> Hm, what's your leaning here?

My preference is to rely on 11ty's instance. The downside is that any point it may go down, or be turned off, and 11ty specifically states there is no [uptime guarantee](https://arc.net/l/quote/nfpntwtm). In that scenario, we can simply remove the provided  background image, and fall back to the lower quality ones until we come up with an alternate solution.

mirisuzanne (approved):
Code looks good to me, but I'm not seeing anything in the deploy preview. Confusing.

jgerigmeyer:
> Code looks good to me, but I'm not seeing anything in the deploy preview. Confusing.

@mirisuzanne What does "not seeing anything" mean in this case? Do you see the embedded videos on e.g. https://deploy-preview-485--oddleventy.netlify.app/2017/01/31/css-day/?

@jamesnw I notice that most of the embedded videos on the site are calling the `embed.figure` macro directly, not using `embed.video`. We may want to address that case too?

jamesnw:
@jgerigmeyer  and @mirisuzanne  Before I transition more content over, can you take a look at the [changes to the macro](https://github.com/oddbird/oddleventy/pull/485/files#diff-ae13f149c7fd9531b93b8fcc900759d40d63b7bdfd099ad4cdc528059e12023c) that I made, and make sure it matches what you're thinking?

content/blog/2023/components.md
```diff
@@ -638,7 +638,7 @@ walking us through the steps:
 
 {{ embed.figure(
   data=[{
-    iframe: 'https://www.youtube.com/embed/R4Ri4ft7bXY',
+    youtube: 'R4Ri4ft7bXY',
     title: 'The Good, The Bad, and The Web Components',
```
jgerigmeyer, line 642: Unrelated to this PR, but it looks like `title` is ignored here -- we might as well remove it anywhere `embed.figure` is called unless we want to support it.

jgerigmeyer:
> @jgerigmeyer and @mirisuzanne Before I transition more content over, can you take a look at the [changes to the macro](https://github.com/oddbird/oddleventy/pull/485/files#diff-ae13f149c7fd9531b93b8fcc900759d40d63b7bdfd099ad4cdc528059e12023c) that I made, and make sure it matches what you're thinking?

@jamesnw It looks on the right track to me. After you port the content, we may want to reconsider the usefulness of the separate `video` macro.

mirisuzanne:
I agree - looks like the right approach, and I think we should probably get rid of the video macro. Just use figure everywhere.

jamesnw:
I applied the `span` attribute to the youtube macro, but I'm wondering if I should also add the width and height? 

On https://deploy-preview-485--oddleventy.netlify.app/talks/css-is-rad/ things are slightly out of alignment-
image

The [live site ](https://oddbird.ing/talks/css-is-rad/#title) is slightly more aligned- 
image

A similar alignment issue is on other tiled layouts with a Vimeo link- https://deploy-preview-485--oddleventy.netlify.app/talks/dynamic-css/ and  https://deploy-preview-485--oddleventy.netlify.app/talks/intrinsic-web/

mirisuzanne:
@jamesnw I added support, but also discovered that the alignment issue is due to how vimeo embeds handle pixel rounding (or something like that) - so it depends on the exact viewport width. When I resize, the youtube embeds resize smoothly, while vimeo embeds are a bit jumpy. Usually things align, and then at some sizes the vimeo embeds lose a couple pixels above and below. I don't think that's on our end.

jgerigmeyer (approved):
Looks good! @jamesnw @mirisuzanne is this ready to merge?

PR #491, opened 2024-01-10
Article: Proxy store

jamesnw:
## Steps to test/reproduce
https://deploy-preview-491--oddleventy.netlify.app/2024/01/09/proxy-store/


content/blog/2024/proxy-store.md
```diff
@@ -0,0 +1,174 @@
...
+  inputFormat: 'scss',
+  outputFormat: 'expanded',
+  inputValue: '',
+  compilerHasError: false,
+};
+
+const playgroundState = new Proxy(initialState, {
+  set(state, prop, ...rest) {
+    // Set state first so called functions have access
+    const set = Reflect.set(state, prop, ...rest);
```
jgerigmeyer, line 112: Is there a reason to specify the arguments to `Reflect.set` here? I think I'd expect:

```suggestion
  set(state, prop, value) {
    // Set state first so called functions have access
    const set = Reflect.set(...arguments);
```

content/blog/2024/proxy-store.md
```diff
@@ -0,0 +1,174 @@
...
+
+This isn't a universal solution. If you have an existing library, use it. If you
+find yourself abstracting out things like `watch` or `computed` for your proxy
+state, you are starting down the road to developing your own framework, and it
+might be a good time to pause and see if your application has grown complex
+enough to bring in something more robust.
+
+This also works best when your logic is contained within a single file. If
+you're wanting to share state values across files, you'll likely be happier
+reaching for another solution.
```
jgerigmeyer, line 174: Hm, I'm curious about this advice -- maybe because I initially read it as implying that the Proxy should only be used in a single file where it's defined. I could imagine an app with relatively simple state logic using a Proxy that's defined in e.g. `store.ts` but then interacted with in various places throughout the app.

jgerigmeyer (approved):
Great article!

jerivas (approved):
Very useful article, and it also taught me about `Reflect`

content/blog/2024/proxy-store.md
jamesnw: We used that pattern for types in the playground, but since this pulled out the types, there's no reason here.

content/blog/2024/proxy-store.md
jamesnw: Good point. I updated to clarify that while you can easily set and access the state across files, the side effects are best colocated with the store itself. 

You _could_ add a `addSideEffect(prop, callback)` method to the store, and then store and call those, but I'd probably recommend against it.

content/blog/2024/proxy-store.md
```diff
@@ -0,0 +1,177 @@
...
+state, you are starting down the road to developing your own framework, and it
+might be a good time to pause and see if your application has grown complex
+enough to bring in something more robust.
+
+This pattern does work well if other files want to update or access the state in
+the store. However, you'll likely want to contain side effect logic to the file
+where the state is defined. If you're wanting to define side effects across
+files, you'd need to do that with callbacks, and likely start looking for ways
+to abstract that out. At that point you'll likely be happier reaching for
+another solution.
```
jgerigmeyer, line 177:
```suggestion
This pattern does work well if other files want to update or access the state in
the store. However, you'll likely want to contain side effect logic to the file
where the state is defined. If you're wanting to define side effects across
files, you'd need to do that with callbacks, and ideally start looking for ways
to abstract that out. At that point you'll probably be happier reaching for
another solution.
```

content/blog/2024/proxy-store.md
```diff
@@ -0,0 +1,177 @@
+---
+title: Proxy is what's in store
+sub: You may not need anything more
+author: james
+date: 2024-01-09
```
jgerigmeyer, line 5:
```suggestion
date: 2024-01-12
```

jgerigmeyer (approved):

PR #499, opened 2024-01-25
Add BT conf and CS Day events

mirisuzanne:
## Description

- BT Conf
- CSSDay

## Steps to test/reproduce

- miriam author page
- talks page

### DIFF
--- a/content/_includes/embed.macros.njk
+++ b/content/_includes/embed.macros.njk
@@ -256,7 +256,7 @@ params:
       page,
       'data.events',
       {'venue': venue} if venue else none
-    ) if page else
+    ) | first if page else
     collection | findData(
       'data.events',
       {'venue': venue}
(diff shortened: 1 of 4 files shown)


content/blog/2018/beyondtellerand.md
```diff
@@ -89,6 +89,7 @@ favorites:
 
 {{ embed.from_event(
   collections.all,
+  page=talk,
```
jgerigmeyer, line 92: When more than one event has the same "venue" it breaks our `embed.from_event` macro unless `page` is specified. This fixes it for this case, but maybe we want something more robust?

content/_includes/embed.macros.njk
```diff
@@ -256,7 +256,7 @@ params:
       page,
       'data.events',
       {'venue': venue} if venue else none
-    ) if page else
+    ) | first if page else
```
jgerigmeyer, line 259: Seems like there was a bug in this macro when `page` was set -- `event` was a list, where we just want the first one (equivalent to `findData`).

jgerigmeyer (approved):

PR #508, opened 2024-02-07
Deploy daily via GH action

jgerigmeyer:
- Replaces IFTTT daily deployment with GH action.
- Upgrades to Node v18 and Yarn v4 (using Corepack instead of bundling Yarn)


jerivas (approved):
TIL about corepack. GH Actions look good too 👍

PR #518, opened 2024-02-20
Package importer blog post

jamesnw:
## Description
Adds a how-to on using the new Package importer

Should not be published until after ESM issues are resolved.

## Related Issue(s)
#517 

## Steps to test/reproduce
https://deploy-preview-518--oddleventy.netlify.app/2024/02/21/pkg-importer/


content/blog/2024/pkg-importer.md
```diff
@@ -0,0 +1,198 @@
...
+package importer allows for relative paths within a module, so I tried
+`pkg:bootstrap-icons/font/bootstrap-icons.css`. However, Webpack was resolving
+files ending with `.css` with `css-loader`, which doesn't understand `pkg:`
+URLs.
+
+At this point, I have a few options. I could import the `.scss` file, tweak the
+Webpack config settings for `css-loader` to not try to resolve `pkg:` URLs, or
+leave it as is. Because I knew I had a story in my backlog to address some other
+issues with `bootstrap-icons`, I decided to leave it as is, and address this as
+part of the upcoming story.
```
jamesnw, line 192: @jgerigmeyer This is a somewhat unsatisfying conclusion, but I wasn't able to find a way to get Webpack to ignore @imports that start with `pkg`- any ideas?

content/blog/2024/pkg-importer.md
```diff
@@ -0,0 +1,198 @@
...
+    Stacks of a variety of cardboard and food boxes,
+    some leaning precariously.
+  position: top
+tags:
+  - Article
+  - Sass
+  - Package management
+summary:
+  Enabling `pkg:` URLs is quick and straightforward
+  in projects using Vite or Webpack `sass-loader`,
```
jgerigmeyer, line 18: I'm curious about this line... It seems even _easier_ in projects using the Sass API directly -- if we want to call out Vite or Webpack here, I don't think we should make it sound like they make the process simpler.

content/blog/2024/pkg-importer.md
jgerigmeyer: Hrm, yeah. I think it's okay to note some of the options here, and I might consider removing the "Because I knew I had..." final sentence.

I haven't tried it, but I'm pretty sure `sass-loader` can parse CSS fine -- would it make sense to configure Webpack to run CSS files through sass-loader first, just like it does with `.scss` files? If so, you could amend this paragraph to:

```
At this point, I have a few options. I could import the `.scss` file and compile
it, tweak the Webpack config settings to use `sass-loader` for `.css` files, or
leave it as is.
```

jgerigmeyer (approved):
Great article! I made some minor edits, I think the most notable was switching the order of the two Webpack examples so it starts with the more standard/default approach.

content/blog/2024/pkg-importer.md
jamesnw: Good point, I'll remove that line.

content/blog/2024/pkg-importer.md
jgerigmeyer: I agree, good call.

jgerigmeyer (approved):

PR #520, opened 2024-02-21
Use sass pkg: importer

jgerigmeyer:


mirisuzanne (approved):
lgtm

src/scss/patterns/_lists.scss
```diff
@@ -1,9 +1,5 @@
 @use '../config';
 
-/// # Typography
-/// Basic typographic styles for the site.
-/// @group type
-
```
jgerigmeyer, line 6: Not sure why, but using the node API caused SassDoc to pick up this file after the `type.scss` file, so this header wasn't at the top of the page.

jamesnw: I'm guessing that has something to do with the change to `inDir` in `compile-docs`, and some odd globbing rule. Did it go from sorting by filename to pathname?

scripts/compile-sass.js
```diff
@@ -1,91 +1,39 @@
...
+import { initAsyncCompiler, NodePackageImporter } from 'sass-embedded';
 
 const __dirname = import.meta.dirname;
 const baseDir = resolve(__dirname, '..');
 const inDir = join(baseDir, 'src', 'scss');
 const outDir = join(baseDir, '_built', 'css');
 const destCssDir = join(baseDir, 'assets', 'css');
 
-let pnp;
+const compiler = await initAsyncCompiler();
```
jgerigmeyer, line 17: @jamesnw Average build times were just under 1s with `sass` using either a compiler or top-level methods. After switching to `sass-embedded`, times dropped to ~650ms using top-level methods, then down to ~450ms using `initCompiler`, and now down to ~300ms using `initAsyncCompiler`.

stacyk (approved):
Looks and works as I'd expect it to. 👍🏻

PR #550, opened 2024-03-27
Address large header-anchor SVGs in RSS feed.

jgerigmeyer:
This PR:

- strips permalinks from markdown headers in RSS feed only
- adds height/width attributes to all svgs

> Hello hello!
>
> First and foremost, thank you for all of the good things you do for the web; I read all of it and I've learned quite a bit over the years.
>
> I use NetNewsWire.app and when I come across an Oddbird post, the link icons look like this (attached). I'm not sure if this is better fixed by the folks at NetNewsWire, or if it's something that could be adjusted on your end — perhaps that SVG needs a height and width defined? Either way, no biggy, but I thought I'd mention it in case others see it as well.

Screenshot 2024-03-26 at 6 58 31 PM


jgerigmeyer:
@mirisuzanne This is ready for review. I don't know a lot about SVGs, so let me know if any of the height/widths I added don't make sense.

src/scss/initial/_headings.scss
```diff
@@ -37,7 +37,7 @@ h3 {
       var(--heading-start),
       var(--heading-end)
     );
-    -webkit-background-clip: text;
+    background-clip: text;
     background-clip: text;
```
mirisuzanne, line 41: This seems like a mistake? Did you update a linter?

src/scss/layout/_main.scss
```diff
@@ -22,7 +22,7 @@
     grid-column: full;
   }
 
-  @media (min-width: 45ch) {
+  @media (width >= 45ch) {
```
mirisuzanne, line 25: Similar, this is maybe an ok change - but it reduces backwards compat. Is a linter doing it, or is there a reason we want to?

mirisuzanne (commented):
Changes to SVGs look good. Less convinced by the changes to CSS. I didn't have a chance to do a full test with an RSS reader yet…

src/scss/layout/_main.scss
jgerigmeyer: Yeah, this seems to be automatically changed by Stylelint: https://stylelint.io/user-guide/rules/media-feature-range-notation/

src/scss/initial/_headings.scss
jgerigmeyer: Hrm, yeah -- this was changed by Stylelint: https://stylelint.io/user-guide/rules/property-no-vendor-prefix/

jgerigmeyer:
> Changes to SVGs look good. Less convinced by the changes to CSS. I didn't have a chance to do a full test with an RSS reader yet…

@mirisuzanne Yeah, good catch. I disabled those Stylelint changes for now, but let me know or feel free to disable or alter the rules globally.

src/filters/type.js
```diff
@@ -108,6 +115,49 @@ export const elide = (html, count = 50) => {
...
+        'link',
+        'listing',
+        'meta',
+        'noscript',
+        'object',
+        'plaintext',
+        'script',
+        'xmp',
+      ].includes(node.tag);
+      const isCustom = ['is-land', 'lite-youtube'].includes(node.tag);
```
jgerigmeyer, line 148: @mirisuzanne After running this through the [W3C validator](https://validator.w3.org/feed/check.cgi?url=https://deploy-preview-550--oddleventy.netlify.app/feed.atom#l799), I figured we probably want to strip out these unsafe and custom tags too. I made the change in https://github.com/oddbird/oddleventy/pull/550/commits/82890cb3d63379ae0cd91e79328957e62ee068d8 -- what do you think?

mirisuzanne: That seems fine to me. I guess check the build output to see if anything surprising is missing?

jgerigmeyer: Out of `feed.atom` looks fine to me -- just with anchor links, a few <script> tags, and custom elements removed.

@mirisuzanne I think this is ready to merge once you've given final review?

jgerigmeyer: @mirisuzanne Do you think this is good to merge?

mirisuzanne: yep, lgtm

mirisuzanne (approved):

PR #567, opened 2024-04-18
Use better links for Sass playground

jgerigmeyer:
- Update Sass playground links to use example
- Add site link to Metecho
- Add side and playground links to Sass Site


stacyk (approved):
I was just hanging around so I thought I would approve if helpful :)

PR #587, opened 2024-05-10
fix content overflow issues on posts

### LINKED ISSUE #583: The page is overflowing at certain screen sizes
There seems to be a breakpoint that triggers a layout change before there is enough space for that change. The result is significant overflow at some intermediate sizes.

Screenshot of overflowing content

- [x] adjust layout @dvdherron 
- [x] review @stacyk 
- [x] review @mirisuzanne 
- [x] apply edits from review @dvdherron 
- [ ] re-review @mirisuzanne 
- [ ] re-review @stacyk

dvdherron:
## Related Issue(s)
fixes #583 
_Reminder to add related issue(s), if available._

## Steps to test/reproduce
View a post with code examples and change the size of the screen from narrow to medium size

## Show me

https://github.com/oddbird/oddleventy/assets/41588129/7c2cc7e0-7584-47e1-9a27-3cd7cc41f978


src/scss/initial/_scale.scss
```diff
@@ -22,7 +22,7 @@ html {
   @include config.above('wide-page') {
     --page-columns: [full-start] var(--page-margin) [content-start]
       minmax(0, 1fr) [wide-start] minmax(0, var(--extra)) [page-start]
-      minmax(min-content, var(--page)) [page-end] minmax(0, var(--extra))
+      minmax(var(--min-page), var(--page)) [page-end] minmax(0, var(--extra))
```
mirisuzanne, line 25: It looks to me like `--min-page` has some other purpose related to auto-fit grids (though the name is pretty unclear). I'm not sure the history of that, but wonder if we want to clarify the name somehow, or separate the different uses into different variables?

This specific overflow seems to be caused by `pre` elements having a `min-content` size based on their full line length, even though they are allowed to overflow. We could address that specific issue by adding `contain: inline-size` on `pre` elements - that might be worth doing in addition to changing the column `minmax()`?

mirisuzanne (commented):
This works but raises some questions for me. :) Thoughts?

src/scss/initial/_scale.scss
dvdherron: > pose related to auto-fit grids (though the name is pretty unclear). I'm not sure the history of that, but wonder if we want to clarify the name somehow, or separate the different uses into different variables?

@mirisuzanne Would something like `grid-single-column` make sense? Or `grid-split-column`? Or just `column`?

Then we would rename `min-page` to whatever we choose above? 

We seem to be using `--min-page` mostly when content is in a single column or split into two columns.

mirisuzanne: Yeah, I think something like that could make sense. Curious if @stacyk has thoughts on that.

mirisuzanne (approved):

src/scss/initial/_scale.scss
stacyk: 1. I don't know if `column` would conflict with `[data-typeset]>* { grid-column: var(--column, page); }`... we use `--column` in many places currently.

2. If it is sometimes used as a single column, but sometimes as a split, then choosing one or the other means it still might not be clear? That said, maybe that is where the difference is?

When it is used on the birds or open source tools, that is using the 22rem size as the column min size. Use it is used in the --page-columns, if re-define `--min-page: 0` I don't see any difference. Where do we need the min width here? 

I can see splitting these out and using a `--column-sm: 22rem;` or something for the data-grids and then use a different variable, if one is even needed, for the --page-columns. 

I could be missing something in the trial of variables though that is not obvious to me.

dvdherron: > 1. I don't know if `column` would conflict with `[data-typeset]>* { grid-column: var(--column, page); }`... we use `--column` in many places currently.
> 
>     2. If it is sometimes used as a single column, but sometimes as a split, then choosing one or the other means it still might not be clear? That said, maybe that is where the difference is?
> 
> 
> When it is used on the birds or open source tools, that is using the 22rem size as the column min size. When it is used in the `--page-columns`, if we re-define `--min-page: 0` I don't see any difference. Where do we need the min width here?

Now that we are containing the size of `pre` elements, replacing `--min-page` with 0 or min-content makes the the columns on posts with `pre` elements more narrow at medium page sizes. 

without a min-width:

bild

with min-width at 22rem:

bild

> 
> I can see splitting these out and using a `--column-sm: 22rem;` or something for the data-grids and then use a different variable, if one is even needed, for the --page-columns.

I went with this route and kept the `--min-page` variable. If this doesn't clarify things, maybe it would make sense to change `--min-page` to `--min-page-column`?

Either way, this PR might need some more work. I just realized that containing the size of `pre` elements is collapsing the cascading color palettes: 

bild

I changed the input column to be 15ch instead of max-content for now but we might want to finesse that some more.

bild

changes in https://github.com/oddbird/oddleventy/pull/587/commits/976d07ca87f35376feadb19b81f6d1a7b0c38fcd

and https://github.com/oddbird/oddleventy/pull/587/commits/9b7bb89dc298898d5bc5dc0fe5778979f619fdac

mirisuzanne: Huh, I wonder if we need that `pre` wrapper on the variable names. But also, we could be more specific about targeting `pre[class|=language]` for containment?

I think `--min-page-column` is pretty accurate to how we're using it here.

I agree that it doesn't _seem_ like we should need it here (we never actually hit that minimum) - but that it _helps_ give the main content some extra priority in the grid layout. I don't see a better way to make the main content expand _before_ the `--extra` space expands.

dvdherron: Sounds good. 

I made the switch to `--min-page-column`, made the `contain: inline-size` change more specific, and reverted the change to the cascading colors variables column in https://github.com/oddbird/oddleventy/pull/587/commits/1968efe3c09b4be6828e5824ae5c478f76301592. I think this is ready for a final review @mirisuzanne (cc @stacyk )

mirisuzanne (approved):

stacyk (approved):
changes look good @dvdherron 🎉

Issue #638, opened 2024-06-19
Add GitHub sponsor badges along with OC sponsors


mirisuzanne:
Probably start with seeing what the GH api exposes, and whether we can merge that in with the data from Open Collective.

PR #639, opened 2024-06-21
Add Github Sponsors

jamesnw:
## Description
Adds a Github Sponsors section

## Related Issue(s)
#638 

image

- [x] Figure out auth
- [x] Put Github levels into Common Loons


content/_data/opencollective.js
```diff
@@ -2,48 +2,106 @@
 import eleventyFetch from '@11ty/eleventy-fetch';
 import { groupBy } from 'lodash-es';
 
+const GITHUBKEY = '';
```
jgerigmeyer, line 5: @jamesnw I think we can add an env var via the Netlify UI (https://app.netlify.com/sites/oddleventy/configuration/env#environment-variables), and then check for it here with `process.env.GITHUBKEY` or whatever. Are there security concerns with that approach?

jamesnw: @jgerigmeyer  I think it's fine to manage the env var that way, but am wondering about the token itself. It sounds like the ideal would be to create a Github App, and then [generate a JWT token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app) to use as the key (which means it isn't tied to a user). Alternatively, we could generate a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) tied to a user's account.

jgerigmeyer: I'd lean toward a PAT, but would defer to @jerivas on this.

content/_data/opencollective.js
```diff
@@ -2,48 +2,106 @@
...
+        query: `
+        {
+  organization(login: "oddbird") {
+    name
+    sponsors(first: 100) {
+      totalCount pageInfo { hasNextPage }
+      nodes {
+        __typename
+        ... on User { name avatarUrl websiteUrl url }
+        ... on Organization { name avatarUrl websiteUrl url }
```
jerivas, line 43: I don't have access to the OddBird org to test this query, but from the [Explorer](https://docs.github.com/en/graphql/overview/explorer) it looks like this should be valid and get us the info we need:

```graphql
{
  organization(login: "oddbird") {
    sponsors {
      nodes {
        ... on User {
          sponsorshipForViewerAsSponsor {
            isOneTimePayment
            isActive
            tier {
              id
              description
            }
          }
        }
        ... on Organization {
          sponsorshipForViewerAsSponsor {
            isOneTimePayment
            isActive
            createdAt
            tier {
              id
              description
            }
          }
        }
      }
    }
  }
}
```

content/_data/opencollective.js
jerivas: I don't feel strongly about this, but JWT seems like the "correct" way. Ideally, we would have a GH application installed on the OddBird org to get this information. However, if that is too annoying, a PAT for our OddBot user feels adequate as well.

content/_data/opencollective.js
jamesnw: @jerivas `sponsorshipForViewerAsSponsor` won't work- the `viewer` is the token's owner, so if we use OddBird's token, it would be "show me sponsorships where OddBird is sponsoring OddBird".

jamesnw: Here's my full running query. I'm still getting no results, and am unsure in `sponsorshipForViewerAsSponsorable` if the `viewer` is "jamesnw" (since I own the token I'm using for testing) or "oddbird". If it's the token owner, I think we'd need to set up the App, since I don't think you can have a Personal Access Token for an organization.

```
{
  organization(login: "oddbird") {
    sponsors(first: 100) {
      nodes {
        ... on User {
          sponsorshipForViewerAsSponsorable(activeOnly: false) {
            isOneTimePayment
            isActive
            tier {
              id
              description
            }
          }
        }
        ... on Organization {
          sponsorshipForViewerAsSponsorable(activeOnly:false) {
            isOneTimePayment
            isActive
            createdAt
            tier {
              id
              description
            }
          }
        }
      }
    }
  }
}
```

jgerigmeyer: I'm fine going the GH App route if that's the right way. [This answer](https://github.com/orgs/community/discussions/24649#discussioncomment-7766153) mentions using sponsorshipsAsMaintainer: https://github.com/xmonad/xmonad-web/blob/0407056255cc253da74a586ef80c4ed16c51d75e/_sponsors.sh

jamesnw: It ended up being `sponsorshipsAsMaintainer` with `isActive:false`. The one Github sponsor was a one-time sponsor. They donated on May 22, and so they were active for exactly a month, and dropped off the list on Saturday.

content/_data/opencollective.js
jamesnw: Let's go with a PAT for our OddBot user. Unless someone is interested in going through the process of setting up the app for the sake of learning, I don't think it's worth the trouble. @jgerigmeyer Can you either put the OddBot login in the shared vault, or create a PAT and replace the existing GITHUB_KEY in Netlify?

content/_data/sponsors.js
```diff
@@ -0,0 +1,132 @@
...
+
+const getDefaultAvatarUrl = (url) => {
+  const slug = url.split('/').at(-1);
+  return slug ? `https://images.opencollective.com/${slug}/avatar.png` : null;
+};
+
+const loadGithubSponsors = async () => {
+  if (!GITHUB_KEY) {
+    // eslint-disable-next-line no-console
+    console.error('Github sponsors not loaded- set GITHUB_KEY in .env.');
```
jamesnw, line 28: I don't feel like this is a requirement for local dev, so opted to put the usage directions here on failure as opposed to in a README.

content/_data/opencollective.js
jgerigmeyer: Moved login info to 1Password 👍

jamesnw: Created the key with odd-bot, updated in Netlify, and added a calendar event to remind us to regenerate in a year.

content/_data/sponsors.js
```diff
@@ -0,0 +1,132 @@
...
+  if (data?.organization.sponsorshipsAsMaintainer.pageInfo.hasNextPage) {
+    // eslint-disable-next-line no-console
+    console.error(
+      'Good news- we have over 100 GitHub sponsors and need to implement pagination.',
+    );
+  }
+  return data?.organization.sponsorshipsAsMaintainer.nodes.map(
+    ({ sponsorEntity }) => ({
+      name: sponsorEntity.name,
+      tier: 'Common Loon',
```
jerivas, line 70: Will this put everyone from GH in the same tier? The [sponsor page](https://github.com/sponsors/oddbird) says

> your avatar [will be] displayed before and larger than lower level sponsors on all [OddTools pages on our website](https://www.oddbird.net/tools/)

jamesnw: Yeah, for now. I'm thinking it'll be simpler to revisit it once we actually have recurring sponsors so that we can code against actual data.

jgerigmeyer (approved):

content/_data/sponsors.js
jgerigmeyer: Yeah, it's a bummer if we can't get tiers _or_ total contributed, but this is fine for now 👍

jamesnw: I think it's likely that we will be able to, but given the uncertainty over naming and data types, if I try to guess at what's being returned now, I'm likely to be wrong and need to redo it anyways.

mirisuzanne (approved):
looks good!

content/_data/sponsors.js
jerivas: I thought the `tiers` field looked promising, but it returned `null` for me and seems like that's a [known issue](https://github.com/antfu-collective/sponsorkit/issues/60) 😕

jamesnw:
@jgerigmeyer Should I be concerned that the github user is showing up in the [deploy preview](https://deploy-preview-639--oddleventy.netlify.app/oddcontrast/)  but not in the [diff](https://github.com/oddbird/oddleventy-built/compare/main..github-sponsors#diff-01b7fcd809fab27dd0d26528ecc14acdeba1c2524c8b9a6807fe6302778b49d9)?

jgerigmeyer:
> @jgerigmeyer Should I be concerned that the github user is showing up in the [deploy preview](https://deploy-preview-639--oddleventy.netlify.app/oddcontrast/) but not in the [diff](https://github.com/oddbird/oddleventy-built/compare/main..github-sponsors#diff-01b7fcd809fab27dd0d26528ecc14acdeba1c2524c8b9a6807fe6302778b49d9)?

No -- the diff is generated without access to the env var, so wouldn't get getting any GH sponsors.

PR #661, opened 2024-07-10
Fetch Mail archives on build

### LINKED ISSUE #655: Build the mail archive on the server

jamesnw:
## Description
Fetches mail archives on build, so that it isn't blocked by user's ad block.

## Related Issue(s)
Closes #655 

## Steps to test/reproduce
Visit https://deploy-preview-661--oddleventy.netlify.app/oddnews/, and see that there are no network requests to https://oddbird.us19.list-manage.com


jamesnw:
@jgerigmeyer I'm having some trouble escaping this properly. It comes in as JS src,  `document.write(<div class=\"with-escapes\"...)`. Because we're on the server, we don't have access to `document`. I am currently stripping out the `document.write("` (and end) but the backslashes remain. 

When I insert it into the template using `{{ newsletterArchives | safe }}`, it looks like this-

image

I've tried `unescape` and `_.unescape`. I've also tried replacing all the `\` characters, which is brittle and emoji still don't work. 

Any ideas on what else to try here?

jgerigmeyer (approved):
@jamesnw Running it through `JSON.parse()` does the trick for removing those string-escaping backslashes.

jamesnw:
@jgerigmeyer Ahhh, I had tried JSON.parse, but without the wrapping quotes. So close!

I realized we were removing and then adding the wrapping quotes back in, so I tweaked the `replace`s to keep them.

PR #685, opened 2024-08-12
Add Favicon link

jamesnw:
## Description
I noticed our logo was not appearing in Google search results.

image

 While we are serving a favicon at `https://oddbird.net/favicon.ico`, there isn't a `link` in the header. Per [Google's docs](https://developers.google.com/search/docs/appearance/favicon-in-search), we should have a link.

## Steps to test/reproduce
Verify the favicon still works. After deployment, we should check Google Search results and verify it works.

PR #694, opened 2024-08-26
Add a lot more favicon options

jamesnw:
## Description
Our favicon still isn't showing up in Google Searches, so I added a more comprehensive set using https://realfavicongenerator.net/.

## Related Issue(s)
#685


src/images/favicon/safari-pinned-tab.svg
jgerigmeyer, line 1: 🙃

jgerigmeyer (approved):

src/images/favicon/safari-pinned-tab.svg
jamesnw: Yeah, I'm confused why it got generated like this. It seemed to work... but also seems to work without the transform, so I removed it.

PR #753, opened 2024-11-21
Winging It 13

SondraE:
## Related Issue(s)
https://github.com/oddbird/oddFlock/issues/178

## Steps to test/reproduce
- Winging It 13 post


SondraE:
@jamesnw 
Please feel free to update the exact wording of _What we cover_ as well as the names of the links. Are there any additional links or demos to add? After you make the updates here, I can also add the same show notes to the video description on YT.

jamesnw:
@SondraE I merged the What we cover and Links section, as the headings give helpful context, I think. What do you think of this strategy?

jgerigmeyer (approved):

SondraE:
@jamesnw 
> I merged the What we cover and Links section, as the headings give helpful context, I think. What do you think of this strategy?

I like this idea.
The spacing between things makes it a bit difficult to see at a glance which content is related, I think.
For example, _What is Baseline?_ and _How are they used?_ are close together but not part of the same point whereas _David Darnes_ and _Keeping an eye_ are farther apart, but part of the same point. Do you have ideas for links or sub points that could go with each bullet point to help make skimming easier? Or maybe we just adjust the spacing a bit?

PR #735, opened 2024-11-04
Remove workshop CTAs

mirisuzanne:
## Description

The workshop has ended, long live the workshop.

PR #763, opened 2024-12-10
Modern CSS workshop

mirisuzanne:
## Description

A trimmed down workshop, focusing on new features.

Should I be covering more of the new stuff, somehow? This still doesn't include:

- anchor positioning
- any animation/transition stuff (view transitions, scroll-based animations, discrete animations, transition-to-auto, starting-style, etc)
- `@property` (I'd have to get into custom properties)
- any of the color stuff (`light-dark()` and `color-mix()` especially)


SondraE (approved):
The _Smashing Conference_ link goes to the workshop page on OddSite. Is that because Smashing doesn't have a page on their site yet for this workshop?

Are you planning to update the image? I would recommend a different one so that it's clear at a glance that this is a different workshop than _Cascading Style Systems_.

mirisuzanne:
@SondraE 

> The _Smashing Conference_ link goes to the workshop page on OddSite. Is that because Smashing doesn't have a page on their site yet for this workshop?

Yes, that's right. We can add the link once it exists.

> Are you planning to update the image? I would recommend a different one so that it's clear at a glance that this is a different workshop than _Cascading Style Systems_.

See what you think.

SondraE:
@mirisuzanne I really like that new image! It's really eye-catching.

content/workshops/modern-css.md
```diff
@@ -0,0 +1,191 @@
+---
+feature: workshop
```
jgerigmeyer, line 2: @mirisuzanne Does this do anything until we revert #735 (and parts of #736 if we want)? Or maybe we want to feature our own April workshop?

mirisuzanne: Oh, yeah like it doesn't do anything currently. I just moved it over from the previous one.

jgerigmeyer: I guess the question is whether we want to feature a workshop, or which one we want to feature next? But for now this doesn't hurt anything.

PR #764, opened 2024-12-10
Minify wash images, update font import

jamesnw:
## Description
In looking at our Lighthouse scores, I found a couple small wins. 

1. Change wash images to webp format. This reduced the size significantly. (178kb to 13kb, 45kb to 6kb)
2. Update how the web font is getting loaded. According to the [embed guide](https://helpx.adobe.com/fonts/using/embed-codes.html#Changingtheembedcodeinawebsite), we were using a legacy mode. Moving it to an import in our CSS file increased our Lighthouse score from 76 to 96. 

I think the improvement was because how the font was being loaded caused a repaint and impacted the initial Largest Contentful Paint.

## Steps to test/reproduce
Check that the images and fonts load as expected.
Run Lighthouse if you are curious. (This only impacts Performance- the SEO and Best Practices ones are wonky due to the Netlify enviroment, I think)

https://deploy-preview-764--oddleventy.netlify.app/


content/_layouts/base.njk
```diff
@@ -7,6 +7,7 @@
   data-ccs-colors>
   <head>
     <meta charset="utf-8" />
+    <meta name="theme-color" content="#4285f4" />
```
jgerigmeyer, line 10: I'm curious where this color comes from?

jgerigmeyer (approved):

content/_layouts/base.njk
jamesnw: Removed. 🤦

PR #798, opened 2025-02-06
Add position-area article

jamesnw:
## Description
Realized I had an article half-written as I investigated `position-area`, so I fleshed it out.

## Related Issue(s)
#797 

https://deploy-preview-798--oddleventy.netlify.app/2025/02/12/anchor-position-area/


content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,170 @@
...
+the different values for positioning an element within that grid.
+
+But `position-area` is more than a shorthand for `anchor()`, introducing some
+behaviors that may be surprising.
+
+## A new containing block
+
+One of the key differences between using `anchor()` and `position-area` is that
+`position-area` creates a new containing block for the positioned element. While
+we're used to dealing with containing blocks when positioning `position:
```
mirisuzanne, line 61: Might be good to say or show something brief about how absolute positioning contexts are created by an element. I imagine many people are used to dealing with it, but not everyone - and it would set up the difference.

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,170 @@
...
+absolute` inside of a `position: relative` ancestor, `position-area` creates a
+containing block that does not map to an element.
```
mirisuzanne, line 63: Does not map to a single element, but the relationship between two elements?

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,170 @@
...
+
+This containing block is generated by the grid area that you select using
+`position-area`. For instance, with `position-area: top left`, the new
+containing block starts at the top left corner of the anchor's containing block
+and extends to the top left corner of the anchor itself.
+
+{{ embed.img(
```
mirisuzanne, line 70: Having a version of this diagram above for absolute positioning could establish that the entire space is the original containing block - and then this one shows the new containing block.

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,170 @@
...
+  src='blog/2025/anchor-area-containing-block.png',
+  alt='A 9-space grid, like a tic-tac-toe board, with an Anchor in the center square, and a square within the top left square for a new containing block.',
+  attrs={
+    'class': 'align-center',
+    'width': '1089',
+    'height': '997'
+  }
+) }}
+
+Note that this is not the positioned element's box, but rather the containing block _for_ the positioned element's box.
```
mirisuzanne, line 80: This feels a bit lingo-y. Is there a way to restate it? Something like… 'Note that the element we're positioning doesn't have to fill this entire area, but we're choosing the container that we want to position our element inside of.'

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Percentages
+
+One of the ways we observe a containing block is that percentages are resolved
+in relation to the containing block. That means a `padding: 5%` on an element
+positioned using `position-area` will have a padding of 5% of the size of the
+containing block, the section of the grid that was selected.
+
+{{ embed.codepen(
```
mirisuzanne, line 89: Helps if these two boxes align and have the same size. I think you can achieve the former by giving the anchors an explicit width. I'm not sure how to make sure they both stretch the same amount. 

Another way to more explicitly visualize the padding area is to layer two backgrounds, and then use background-clip so that the top one is clipped at the content-box.

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,170 @@
...
+In this case, the padding on the element positioned with `anchor()` is much
+larger, since its containing block is the viewport, not the grid area created by
+`position-area.
+
+## Self alignment
+
+Using `position-area` also changes the normal `self-alignment`, so that the
+positioned element is placed near to the anchor within its containing block.
+
+What that means is that you can use `place-self` on the positioned element, even
```
mirisuzanne, line 105: I would put 'self-alignment' in quotes rather than ticks, since it's not a property. 

There are two things happening here that feel a bit muddled to me. One is that we need to opt out of `auto` insets for self-alignment to work. We can do that in the function example by setting `top` and `left` insets to `0` - so we have the full 'inset-modified' positioning context to align against. Selecting an area does that for us automatically, which allows us to `place-self` without setting explicit inset values. The other thing that's happening is we get a _default alignment_ next to the anchor. Does it make sense to think of those as different things?

All of that behavior can be recreated with the anchor function, but it requires more code.

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,170 @@
...
+  title='position-area self alignment - absolute positioning',
+  user='jamessw',
+  height=350
+) }}
+
+Note that in that example, the targets are `position: absolute`. If instead, the
+targets are `position: fixed`, the targets do not overflow their original
+containing block (the viewport) and so no shifting happens.
+
+{{ embed.codepen(
```
mirisuzanne, line 141: If I recreate the effect of `position-area: center` manually…

```css
  left: anchor(left);
  right: anchor(right);
  top: anchor(top);
  bottom: anchor(bottom);
  place-self: center;
```

…I get the same shifting behavior. So I don't think this is a specific feature of position-area, but a feature of how positioning, self-alignment, and overflow behave generally? Here it is again, without using anchors at all: https://codepen.io/miriamsuzanne/pen/xbKvKPg?editors=1100

mirisuzanne (commented):
Nice work! I enjoyed this overall, but a few concepts maybe need to be pulled apart more - between how positioning containers are established, and how alignment works within those containers.

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+behaviors that may be surprising.
+
+## A new containing block
+
+One of the key differences between using `anchor()` and `position-area` is that
+`position-area` creates a new containing block for the positioned element.
+
+If you are using `anchor()`, the positioned element's containing block works
+like normal. For a `position: fixed` positioned element, it will usually be the
+viewport. For a `position: absolute` positioned element, it will usually be the
```
mirisuzanne, line 72:
```suggestion
like normal. For a `fixed` position element, it will usually be the
viewport. For an `absolute` positioned element, it will usually be the
```

I don't know if this is an improvement, it just reads a bit easier to me. But take it or leave it.

content/blog/2025/anchor-position-area.md
mirisuzanne: If there's an anchor relationship, there should still be a relationship, right? Or am I missing something? Is the position-area value not always an inset fragment of the containing block that `anchor()` would reference?

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+```css
+.positioned-element {
+  bottom: anchor(top);
+  left: anchor(right);
+  top: 0;
+  right: 0;
+  justify-self: start;
+  align-self: end;
+}
+```
```
mirisuzanne, line 125: This comparison works really well!

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Percentages
+
+One of the ways we observe a containing block is that percentages are resolved
+in relation to the containing block. That means a `padding: 5%` on an element
+positioned using `position-area` will have a padding of 5% of the size of the
+containing block, the section of the grid that was selected.
+
+{{ embed.codepen(
+  id='MYgMpMM',
```
mirisuzanne, line 180: I think equal-size anchors would still help here, but not essential to see the difference.

mirisuzanne (approved):
This is great - the changes work real well!

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+element to be touching a specific corner or edge of the anchor. `position-area`
+creates a grid around the anchor, and lets you place your positioned element
+within that grid.
+
+[Anchor-tool.com](https://anchor-tool.com/) is a fantastic tool for exploring
+the different values for positioning an element within that grid.
+
+[{{ embed.img(
+  src='blog/2025/anchor-tool-com.jpg',
+  alt='A screenshot of anchor-tool.com, demonstrating an element positioned with postion-area: top span-left.',
```
jgerigmeyer, line 56:
```suggestion
  alt='A screenshot of anchor-tool.com, demonstrating an element positioned with position-area: top span-left.',
```

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+As a side effect of being an absolutely positioned element with `justify-self`
+or `align-self` set, your positioned element will have the behavior of smartly
+shifting to stay within its original content block.
+
+If, after applying padding and margins, your positioned element would overflow
+the newly created containing block but would fit inside its original container
+block, it will get shifted to prevent overflow.
+
+In this example, the targets are all positioned with `position-area: center`.
+The top and bottom targets are shifted towards the center because they would
```
jgerigmeyer, line 145:
```suggestion
The top and bottom targets are shifted toward the center because they would
```

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+{{ embed.codepen(
+  id='ByBgbQa',
+  title='position-area self alignment - fixed position',
+  user='jamessw',
+  height=350
+) }}
+
+You can opt in to this behavior with `anchor()` by setting a value for all of
+`top`, `right`, `bottom`, and `left`, and setting a value for `justify-self` and
+`align-self`. But `postion-area` is much more terse.
```
jgerigmeyer, line 170:
```suggestion
`align-self`. But `position-area` is much more terse.
```

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+{{ embed.codepen(
+  id='XJrLzWa',
+  title='position-area self alignment - absolute positioning',
+  user='jamessw',
+  height=350
+) }}
+
+Note that in that example, the targets are `position: absolute`. If instead, the
+targets are `position: fixed`, the targets do not overflow their original
+containing block (the viewport) and so no shifting happens.
```
jgerigmeyer, line 159:
```suggestion
Note that in this example, the targets are `position: absolute`. If instead the
targets are `position: fixed`, the targets do not overflow their original
containing block (the viewport) and so no shifting happens.
```

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+) }}
+
+You can opt in to this behavior with `anchor()` by setting a value for all of
+`top`, `right`, `bottom`, and `left`, and setting a value for `justify-self` and
+`align-self`. But `postion-area` is much more terse.
+
+## Percentages
+
+One of the ways we observe a containing block is that percentages are resolved
+in relation to the containing block. That means a `padding: 5%` on an element
```
jgerigmeyer, line 175:
```suggestion
in relation to the containing block. This means that `padding: 5%` on an element
```

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+positioned using `position-area` will have a padding of 5% of the size of the
+containing block, the section of the grid that was selected.
```
jgerigmeyer, line 177:
```suggestion
containing block -- the section of the grid that was selected.
```

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+{{ embed.codepen(
+  id='MYgMpMM',
+  title='position-area containing block',
+  user='jamessw',
+  height=500
+) }}
+
+In this case, the padding on the element positioned with `anchor()` is much
+larger, since its containing block is the viewport, not the grid area created by
+`position-area.
```
jgerigmeyer, line 188:
```suggestion
`position-area`.
```

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Try it out
+
+`position-area` is available in Chromium browsers. It was initially called
+`inset-area`, but was renamed in the Chromium 129 release.
+
+The [Anchor Positioning Polyfill](https://anchor-polyfill.netlify.app/) does not
+yet support `position-area`. This article was written as part of our exploration
+into the spec, and if you would like to sponsor further development, please get
+in [touch](/contact/)!
```
jgerigmeyer, line 198:
```suggestion
into the spec, and if you would like to sponsor further development, please
[get in touch](/contact/)!
```

jgerigmeyer (approved):
👏

content/blog/2025/anchor-position-area.md
jamesnw: Oh right, thanks for the reminder!

content/blog/2025/anchor-position-area.md
jamesnw: For positioned elements with absolute positioning, the containing block can be an ancestor of the anchor's containing block. If you're using `anchor()`, the positioned element isn't contained by the anchor's containing block, so there wouldn't be a direct relationship. Or I might be missing your point here?

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+
+{% callout 'note', false %}
+
+The examples in this article are best viewed in a Chromium browser, version 131
+or later.
+
+{% endcallout %}
+
+On the surface, `position-area` seems like just a shorthand.
+Just replace `top: anchor(bottom); left: anchor(right)`
```
stacyk, line 33: May I suggest we consider replacing one of the "just"s here with another word?
```suggestion
On the surface, `position-area` seems like a shorthand.
Just replace `top: anchor(bottom); left: anchor(right)`
```

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+within that grid.
+
+[Anchor-tool.com](https://anchor-tool.com/) is a fantastic tool for exploring
+the different values for positioning an element within that grid.
+
+[{{ embed.img(
+  src='blog/2025/anchor-tool-com.jpg',
+  alt='A screenshot of anchor-tool.com, demonstrating an element positioned with position-area: top span-left.',
+  attrs={
+    'class': 'align-center'
```
stacyk, line 58:
```suggestion
    'class': 'align-center img-shadow'
```

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+The idea behind `position-area` is that in most cases, you want your positioned
+element to be touching a specific corner or edge of the anchor. `position-area`
+creates a grid around the anchor, and lets you place your positioned element
+within that grid.
+
+[Anchor-tool.com](https://anchor-tool.com/) is a fantastic tool for exploring
+the different values for positioning an element within that grid.
+
+[{{ embed.img(
+  src='blog/2025/anchor-tool-com.jpg',
```
stacyk, line 55: @jamesnw would it be acceptable to replace this image with one cropped to the inner example since I think the white around the edges isn't even? I'm not sure if it is important to have the checkboxes as I actually clicked one thinking it was an embed for a second 😵 oops :) 

anchor-tool

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+  }
+) }}
+
+Note that the element we're positioning doesn't have to fill this entire area,
+but we're choosing the container that we want to position our element inside of.
+
+## Positioning within the containing block
+
+Using `position-area` also handles some defaults so that the positioned element
+is placed near to the anchor, within its new containing block.
```
stacyk, line 102: Do we need the word "to" here? My brain kept wanting to read this as ... 

is placed near the anchor, within ...

stacyk: ah, looked it up after I read the "close to" in a few lines

https://dictionary.cambridge.org/us/grammar/british-grammar/near-and-near-to

content/blog/2025/anchor-position-area.md
```diff
@@ -0,0 +1,209 @@
...
+
+First, any inset properties like `top` or `inline-start` that don't have values
+set on the positioned element resolve to `0`. This makes the containing block
+fill the entirety of its grid area.
+
+Then, `align-self` and `justify-self` are used to position the element close to
+the anchor, by specifying the alignment away from the specified grid area and
```
stacyk, line 109:
```suggestion
the anchor, by setting the alignment away from the specified grid area and
```

since we are using specified and non-specified in this sentence, maybe we can adjust the word "specifying" to something else?

content/blog/2025/anchor-position-area.md
stacyk: Great demo.

stacyk (approved):
I left a few suggestions but nothing I suggested is breaking. I think this will be a great reference for people. 

I was playing around with one of your demos from the last article... what if we had a CodePen template that pulls some CSS settings we can use on all OddSite pens that uses our typeface and brand colors. Is that something that you'd be interested in? 

For instance, swapping any color references from coral and lavender to var(--color-1) or var(--color-2) and then pasting this URL in the Settings / CSS "Add External Stylesheets/Pens" field 

See this example:
https://codepen.io/stacy/pen/wBwVwLL?editors=1100

```
https://codepen.io/stacy/pen/KwKVrMX/0388bd5ed4d398ddda3e26f2fcdd0e1f.css
```

mirisuzanne:
@stacyk it looks like @jamesnw made some revisions after your review? Is this back in your court? Or are we waiting on more from @jamesnw here?

PR #823, opened 2025-03-06
Add Learn With Jason - Anchor Positioning

jamesnw:
## Related Issue(s)
#429


SondraE (approved):
lgtm!
I think we probably want to move the hire us cta block down the page, positioning it just above the "Let's chat" and "OddNews" footer. Having it above the related posts make it seem like it's part of the content of your post. (Perhaps a different PR.)

PR #850, opened 2025-03-18
OddNews Post

SondraE:
## Description
Adding a new Post Type called OddNews as well as Issue 21 that came out March 13, 2025

## Related Issue(s)
[#837](https://github.com/oddbird/oddleventy/issues/837)

## Steps to test/reproduce
- [ ] Add Mia's article from March OddNews as an article post with invite to subscribe to OddNews.
- [ ] Add OSS updates from March OddNews as an article post with tag "oss".

### DIFF
--- a/content/_data/taxonomy.yaml
+++ b/content/_data/taxonomy.yaml
@@ -21,6 +21,9 @@ post:
   - tag: Winging It
     icon: winging-it-lines
     plural: Episodes
+  - tag: OddNews
+    icon: winging-it-lines
+    plural: Issues
 
 oss:
   - title: Open Web
--- a/content/blog/oddnews/oddnews-21.md
+++ b/content/blog/oddnews/oddnews-21.md
@@ -1,11 +1,9 @@
 ---
 title: CSS Cascade Layers
 issue: 21
-author:
-  - birds
+url: http://eepurl.com/i_NTYw
 date: 2025-03-13
 tags:
-  - OddNews
   - CSS
   - Anchor Positioning
   - Typography
@@ -20,7 +18,6 @@ summary: |
   - Custom CSS Functions in the Browser article by Miriam Suzanne
 ---
 
-{% import "embed.macros.njk" as embed %}
 {% import "utility.macros.njk" as utility %}
 
 {{ utility.main_action(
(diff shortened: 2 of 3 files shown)


SondraE:
**Questions**:

- Does someone else have time to make the new Post Type work correctly?
- It feels a bit odd to include things like Winging It and post highlights in this context. Should we limit these OddNews posts to the content that hasn't been published already on OddSite with a note that OddNews subscribers get additional content and bird facts?

content/_data/taxonomy.yaml
```diff
@@ -21,6 +21,9 @@ post:
   - tag: Winging It
     icon: winging-it-lines
     plural: Episodes
+  - tag: OddNews
+    icon: winging-it-lines
```
mirisuzanne, line 25: what if we used something like the oddbird logo for this?

content/blog/oddnews/oddnews-21.md
```diff
@@ -0,0 +1,198 @@
+---
+title: CSS Cascade Layers
```
mirisuzanne, line 2: I guess a lot of this metadata isn't updated yet?

content/blog/oddnews/oddnews-21.md
```diff
@@ -0,0 +1,198 @@
...
+issue: 21
+url: http://eepurl.com/i_NTYw
```
mirisuzanne, line 4: I think `url` is only used if we also have a `venue` which would make sense for these posts.

content/blog/oddnews/oddnews-21.md
```diff
@@ -0,0 +1,198 @@
...
+  - CSS Layout Workshop coming in April 2025
+  - Custom CSS Functions in the Browser article by Miriam Suzanne
+---
+
+{% import "utility.macros.njk" as utility %}
+
+{{ utility.main_action(
+  'Subscribe to OddNews »',
+  subscribe_url
+) }}
```
mirisuzanne, line 26: I see we're using this pattern on all winging-it post, probably so we can get access to the `subscribe_url`? I think we could be using the built-in actions from yaml if we prefer. It would be a bit more consistent with other posts:

```yaml
eleventyComputed:
  action:
    text: Subscribe to OddNews »
    url: '{{ subscribe_url }}'
```

content/blog/oddnews/oddnews-21.md
```diff
@@ -0,0 +1,198 @@
...
+  opaque.
+
+### from [James Stuckey Weber](/author/james/)
+
+I reviewed feature changes for the Web Features project and contributed to the
+Web DX Community Group. I also contributed a bug fix to the Firefox Dev Tools to
+show the offsetParent of fixed position elements if it’s not the viewport. It
+should be included in the Firefox 137 release.
+
+{% callout 'note' %}
```
mirisuzanne, line 61: This doesn't feel right to me as a callout under the 'James' heading. I think it should get it's own section in a blog post like this one.

content/blog/oddnews/oddnews-21.md
```diff
@@ -0,0 +1,198 @@
...
+articles](https://www.bram.us/2025/02/18/css-at-function-and-css-if/) that go
+into more detail. I’ve also been playing in CodePen with more
+[color-manipulation](https://codepen.io/miriamsuzanne/pen/dPyzLEJ) and
+[fluid-type functions](https://codepen.io/miriamsuzanne/pen/ogNobGx).
+
+What functions would be most useful to you? Let us know what you come up with on
+[Mastodon](https://front-end.social/@OddBird) or
+[BlueSky](https://bsky.app/profile/oddbird.dev), we’d love to see your demos.
+
+{% callout 'note' %}
```
mirisuzanne, line 184: Do we have a CTA for this? I think that would be better than copy/pasting this callout.

mirisuzanne (commented):
@SondraE What isn't working about the post type? That part looks fine on my end.

To your other question, I'm curious which goal we're aiming for - because they might lead us different directions. I could imagine us finding a way to duplicate the email basically intact - so this would act as a pretty direct archive. But I could also imagine us splitting things out even farther.

- OSS updates could become their own post
- We don't currently have a way to 'announce' things on the website, like upcoming live streams or workshops, but we've talked before about adding that.

This current PR falls somewhere in-between, combining the announcements with the OSS updates - so it's most of the email, but not the links to old posts. That can work too. But if we're going with a multi-topic post, I think we could go all in and even keep the links to old articles. The '[vibe check](https://daverupert.com/2025/02/vibe-check-37/)' style post is common on personal sites that I follow.

I don't know which is better. In some ways I like giving each type of content a home. In other ways, I like the roundup. 

I suppose the most complex option is a "why not both?" approach - where we start by creating individual oss posts and announcements, and then we also do a monthly summary post with some overarching commentary.

SondraE:
@mirisuzanne 

> What isn't working about the post type? That part looks fine on my end.
Jonny fixed the things that were breaking the build. Turns out it wasn't the post type that I'd messed up. It was something in the template.

> I'm curious which goal we're aiming for.
I like your reflections on the options for what this OddNews post could be. 

I lean heavily toward the OSS updates and original articles being pulled out and published on OddSite. We already have an OddNews archive list on OddSite that links to pages with exact replicas of each newsletter. We should make that archive easier to find.

So maybe we shouldn't go with an _OddNews_ post type. Instead, we use the _Link_ tag for any piece of content from OddNews that we want to republish on OddSite, and the posts include small portions of the original content with a CTA to go read the full article or OSS update in the specific issues of OddNews where it first appeared. We could also include the archive list of past newsletters, perhaps truncated in some way. And we could have a "sign up for OddNews" CTA.

mirisuzanne:
@SondraE 

> So maybe we shouldn't go with an _OddNews_ post type. Instead, we use the _Link_ tag for any piece of content from OddNews that we want to republish on OddSite, and the posts include small portions of the original content with a CTA to go read the full article or OSS update in the specific issues of OddNews where it first appeared. 

Hmmm, that feels to me like we're adding an extra step that would usually go the other direction. It feels strange to write a short post in the email, and then copy an even shorter segment of it onto the blog, and make people click through to an email archive to keep reading. Usually we're sending people from our emails to the website, which feels more natural. If we're duplicating OSS updates on the blog, which I think is good, I think it's nicer to have the full post.

> We could also include the archive list of past newsletters, perhaps truncated in some way. And we could have a "sign up for OddNews" CTA.

We could still link to the newsletter from those posts if we want, and have a signup CTA. The funny thing is that our emails often have a support us CTA, which would also be an option on the blog posts. aaaaaah, so many calls to so many actions, just give us money already! ha

PR #855, opened 2025-03-24
Hotfix signup form image translation

mirisuzanne:
I did the original translation in a hurry, and I think I may have done it badly.


jgerigmeyer (approved):
I'm not sure I'm qualified to review this change? Google Translate says this is a fine change, but also doesn't complain about the original. Is there a different image we could choose, since (afaik) we don't have someone fluent on hand to review?

mirisuzanne:
@jgerigmeyer I can check the translation in the pen with a friend - but how's this for a different image?

jgerigmeyer (approved):

PR #899, opened 2025-05-22
Winging It 19

SondraE:


SondraE:
@stacyk Now that we know what you all covered in the CSS Q&A Party episode, I think we'd get more watches if we changed the title to the most interesting/beneficial thing you all covered. I played around with a subtitle in this PR, but I think we could find something even more appealing. I'm curious to hear your thoughts on new title options.

jgerigmeyer (approved):

content/blog/wingingit/winging-it-19.md
```diff
@@ -0,0 +1,70 @@
+---
+title: What's New to CSS?
```
stacyk, line 2: @SondraE I don't know what is the technically correct way to phrase this, but I would have said `What's New in CSS`. ¯\_(ツ)_/¯

stacyk (approved):

PR #904, opened 2025-05-29
Adjust video course url

dvdherron:
## Description
Change CSS video course url to /learn/courses/design-with-code/

## Related Issue(s)
_Reminder to add related issue(s), if available._

## Steps to test/reproduce
Go to /learn/courses/design-with-code/ or navigate to the course under the Courses tag

## Show me
image

### DIFF
--- a/content/learn/courses/video/advanced-css.md
+++ b/content/learn/courses/video/advanced-css.md
@@ -13,7 +13,7 @@ summary: |
   the fundamentals behind
   each technology.
 author: miriam
-permalink: learn/design-with-code/
+permalink: learn/courses/design-with-code/
 ---
 
 {% import 'layout.macros.njk' as layout %}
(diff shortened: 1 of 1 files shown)


jgerigmeyer (approved):
> @jgerigmeyer (and/or @jamesnw ) @mirisuzanne Does just changing the permalink here work or is there something else that would also need to change?

@dvdherron I think just changing the permalink like this is fine, so I approved this PR. Two non-blocking questions though:

1. Do we want an image for this course/post?
2. To be consistent with the other course, should this URL be `learn/courses/design-with-code/`?

jgerigmeyer (approved):

SondraE:
@dvdherron @jgerigmeyer @mirisuzanne 

> I'm happy with /learn/courses/design-with-code/

If Mia's happy with that, I'm happy with that!

> I'd also be happy with an image, but don't know how it fits in with the current design patterns or planned changes.

An image would match our current pattern, but because I didn't have an obvious one and wanted to get this out quickly, and because I do have some page layout patterns proposals coming that don't require a pic for every detail page, I decide not to worry about an image here for now.

PR #924, opened 2025-06-23
Add related posts

### LINKED ISSUE #920: Be able to select a tag for related posts (instead of automatically relating by type)

jamesnw:
## Description
Adds a way to add "Related Posts" rather than "Recent Posts".

If a `related_tag` is set, then the most recent posts with a tag are shown. Otherwise, the current "Recent Articles" or "Recent Talks" or "Recent Videos" section will be shown.

**Questions:**

- [x] This is limited to 1 related tag- should this support multiple tags? Not yet
- [x] This shows all post types (Podcast, Talk, Video, Link, Article, etc). This should exclude Talks
- [x] Move element higher

## Related Issue(s)
Closes #920 

## Steps to test/reproduce
View https://deploy-preview-924--oddleventy.netlify.app/2025/05/06/polyfill-updates/, where the 3 most recent items tagged "Anchor Positioning" are shown. 

All other sites still show the "Recent Posts" block instead.

### DIFF
--- a/content/_includes/page/base.njk
+++ b/content/_includes/page/base.njk
@@ -59,13 +59,13 @@
     </section>
   {% endif %}
 
+  {% include 'page/list.njk' %}
+
   {% if cta == 'hire' %}
     {% include 'site/hire-us.njk' %}
   {% elif cta == 'sponsor' -%}
     {% include 'site/sponsor-us.njk' %}
   {% endif %}
-
-  {% include 'page/list.njk' %}
 
   <!-- used for webmentions -->
   <a href="{{ page.url | url }}" class="u-url" hidden>{{ (banner or title) | removeMd }}</a>
--- a/content/_includes/page/list.njk
+++ b/content/_includes/page/list.njk
@@ -44,5 +44,10 @@
     collections.all | getCollectionItem(page) | pageEvents
   ) }}
 {% elif is_entry %}
-  {% include 'page/recents.njk' %}
+  {% if related_tag %}
+    {% include 'page/related.njk' %}
+  {% else %}
+    {% include 'page/recents.njk' %}
+  {% endif %}
+
 {% endif %}
--- a/content/_includes/page/list.njk
+++ b/content/_includes/page/list.njk
@@ -49,5 +49,4 @@
   {% else %}
     {% include 'page/recents.njk' %}
   {% endif %}
-
 {% endif %}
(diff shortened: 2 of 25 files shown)


SondraE (changes requested):
@jamesnw Thanks for taking this one on!

**My Suggestions:**

- In an effort to help more people see the related content, I'd recommend putting this section above _Hire Us_ and probably also above the _Open Source Sponsors_, although, I could be convinced that it should go below sponsors in specific cases. 

- How much of a lift would it be to display tags in the list views? Or, even better, to only display the tag that was used to relate the posts? When I see the cards in the related posts section, it makes me think that it would be useful to know that the posts are related by "anchor positioning." Should "displaying tags in list views" be a separate issue?

> This is limited to 1 related tag- should this support multiple tags?

For now, I think we could start with support for one tag and see whether we want/need to support multiple tags in the future.

> This shows all post types (Podcast, Talk, Video, Link, Article, etc). Should they all be included? Should this be customizable per post?

I recommend excluding Talks/Workshops just as we exclude them from the list of all posts. If there's a timely Talk or Workshop that we want to feature at the bottom of a post, I think that should be handled separately.

I think it would be ideal if we had the option to select three specific posts to use as the related posts for each individual post, but as an incremental improvement, being able to select one tag and include all posts types (except talks/workshops) is a good first step, IMO.

jamesnw:
> * In an effort to help more people see the related content, I'd recommend putting this section above _Hire Us_ and probably also above the _Open Source Sponsors_, although, I could be convinced that it should go below sponsors in specific cases.

This order is set in the same base layout for all pages. Should this be a global change? Or just for some types?

> * How much of a lift would it be to display tags in the list views? Or, even better, to only display the tag that was used to relate the posts? When I see the cards in the related posts section, it makes me think that it would be useful to know that the posts are related by "anchor positioning." Should "displaying tags in list views" be a separate issue?

Is this what you mean? 

image

SondraE:
@jamesnw 

> This order is set in the same base layout for all pages. Should this be a global change?

Yes, I was thinking this should be a global change.

> Is this what you mean?

That is one option I was thinking about. We could even remove the word _Related_.
Another option would be something like this. Is one of these easier to pull off?

Screenshot 2025-06-23 at 3 56 57 PM

SondraE (approved):
LGTM!

jamesnw:
I added `related_tag` to all articles from 2025, and articles from the past getting significant traffic.

jamesnw:
> I find the current wording can sometimes be a bit confusing (e.g. "Testing Posts"), and might lean toward something more verbose (e.g. "Related Posts about <...>"), but I don't feel strongly.

@SondraE  I changed this to the inverted but slightly less verbose "Posts about <...>". Does that work?

PR #965, opened 2025-08-26
Article to go with video and demo

mirisuzanne:
## Description
Uncomment if you want to provide a custom description.

Article that accompanies type video and demo

(still needs an image)


SondraE:
@mirisuzanne @jgerigmeyer 
It's working now for me too.

PR #971, opened 2025-09-02
Split out anchor positioning rules

jamesnw:
## Description
Creates a separate entry point for the anchor related styles. Hoping this prevents a content flash.

## Steps to test/reproduce
Go through different pages in Firefox, and see if there are content flashes.


jamesnw:
@dvdherron Does this fix it? I think I'm still getting a flash, but can't tell as easily.

dvdherron:
> @dvdherron Does this fix it? I think I'm still getting a flash, but can't tell as easily.

This does fix the issue. I didn't realize I had "disable cache" checked under the network tab in dev tools. 

(On the live site, no matter if "disable cache" is checked or not there's still a very noticeable flash.)

dvdherron (approved):

jgerigmeyer (approved):

Issue #970, opened 2025-09-02
Bug: delayed loading of stylesheet causes a brief layout flash/shift whenever a page loads


dvdherron:
- Doesn't happen in Chrome, only Firefox (tested on beta v. 143.0b6)
- If you look in the head after a page load, you can see the style tag with the `data-original-href` attribute load after a delay. 

Could be something to do with the anchor positioning polyfill?

PR #972, opened 2025-09-03
Remove anchor positioning

jamesnw:
## Description
Anchor positioning seems to causing a brief flash on style replacement, and isn't actually needed for our single use case. This removes it from the site.

## Related Issue(s)
#970


stacyk (commented):
@jamesnw hmm, can we position popover relative to the button that activates it without anchor positioning? I see it works in Safari and FF, but not in Chrome.

dvdherron (commented):
I initially thought this still had a subtle flash, but I realized I had "disable cache" checked in the dev tools. With that unchecked, this version with the polyfill removed works as expected. 

But it's a moot point since your changes in #971 seem to fix the issue. 

 (The live site has a noticeable flash no matter if "Disable Cache" is checked or not)

PR #985, opened 2025-09-18
Draft CSS post, and architecture in nav

mirisuzanne:
## Description

There's no reason for these two things to be in the same PR, so we can split them out if it's helpful to merge separately. This is the draft post I wrote yesterday when trying to figure out what to say on socials. Still needs a new image, and maybe a very basic codepen? I don't think there's any rush to get it out, though. We could aim for next week?

### DIFF
--- a/content/_data/site.yaml
+++ b/content/_data/site.yaml
@@ -28,8 +28,8 @@ nav:
         url: /services/planning/
       - title: Design Systems
         url: /services/design-systems/
-      - title: Consulting
-        url: /services/consulting/
+      - title: Poetic CSS Architecture
+        url: /services/architecture/
       - title: Web Standards
         url: /tools/#featured-funding-goal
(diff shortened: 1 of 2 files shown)


mirisuzanne:
See https://github.com/oddbird/oddleventy/pull/986 for just the nav change

SondraE (changes requested):
One typo: "If I want _a_ my calculations..."

Switch the CTA note at the bottom of the article to pitch and link to the poetic architecture services landing page.

**Image Options**
alvaro-serrano-hjwKMkehBco-unsplash
pexels-suzyhazelwood-3928938
pexels-suzyhazelwood-6268028
trust-tru-katsande-wDJrBpl-q3Q-unsplash (1)

jgerigmeyer (approved):

PR #999, opened 2025-10-03
Django Landing Page

SondraE:
## Tasks

- [x] @SondraE PR with Django landing page template
- [ ] @vwallen & @jerivas Write text describing benefits
- [ ] @vwallen Add link from Python-related UV article

## Description
Here is the outline/template for a Django-related services landing page.
All the sections that say "Benefit" and the sections in brackets are placeholder text.

## Related Issue(s)
#998 

## Steps to test/reproduce
/services/django/

### DIFF
--- a/content/services/django.md
+++ b/content/services/django.md
@@ -1,0 +1,310 @@
+---
+title: GET BENEFIT with a Django Fixup
+sub: >
+  BENEFIT ONE,
+  BENEFIT TWO,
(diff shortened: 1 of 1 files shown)


jerivas:
I filled out all sections (except for the video) with a focus on speeding up and securing old Django codebases. I think that's a pain point that many potential clients have and is well suited to our strengths

